PR #11, opened 2023-05-30
MNT: Use '--skip-existing' over '--force'

matthewfeickert:
Use `anaconda-client` `upload` command's `--skip-existing` flag over the `--force` flag to avoid redundant uploads, saving time, and to avoid uploading when legitimate errors exist. From `anaconda upload --help`:
```
...
--force: Force a package upload regardless of errors
--skip-existing: Skip errors on package batch upload if it already exists
...
```

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -38,5 +38,8 @@ env
 # upload wheels
 echo "Uploading wheels to anaconda.org..."
 
-anaconda -t $ANACONDA_TOKEN upload --force -u "$ANACONDA_ORG" "$INPUT_ARTIFACTS_PATH"/*.whl
-echo "Index: https://pypi.anaconda.org/$ANACONDA_ORG/simple"
+anaconda --token "${ANACONDA_TOKEN}" upload \
+  --user "${ANACONDA_ORG}" \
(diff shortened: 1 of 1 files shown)


Carreau:
I'm pretty sure IPython overwrite the same `.dev0` as well, not because I want to, but because I never had the time to properly set it up to actually bump the dev version on nightly build and cleanup the `setup.py` (yet we still have a setup.py, I know, and yes I want to remove it).

matthewfeickert:
Given https://github.com/Anaconda-Platform/anaconda-client/issues/655#issuecomment-1570392111 `anaconda-client` is uploading based off of name, so I'll have to think about if there is some cleaner way to do this. Putting this into draft for the time being, but maybe closing later if I can't think of something better.

stefanv (approved):

Carreau:
I have no objections of having parameters, but at the same time I also do not want to make this a kitchen-sink utilities. At some point it should be ok to say "this is beyond the scope of this action, if you wish to have more control please use XXX". For example https://github.com/OpenAstronomy/publish-wheels-anaconda has more parameters and also take care of the removal.

Issue #8, opened 2023-05-26
Add ability to select package index for upload


matthewfeickert:
At the moment the package index to upload the distributions is hardcoded to https://anaconda.org/scientific-python-nightly-wheels

https://github.com/scientific-python/upload-nightly-action/blob/80af44feb2d73fe29869d46207c02f88885ea5b8/entrypoint.sh#L17

Related to https://github.com/scientific-python/upload-nightly-action/issues/5#issuecomment-1564723037, this should not be hardcoded as there are additional nightly package indexes that can reasonably be desired to be used. As an explicit example, at the moment `matplotlib` uploads nightly wheels to https://anaconda.org/scipy-wheels-nightly (@tacaswell @QuLogic I'm not sure if there was discussion at the 2023 summit about switching where you do nightly uploads to, so I'll let you comment here).

---

You may be very reasonably wondering why there is an https://anaconda.org/scientific-python-nightly-wheels/matplotlib if that's not where `matplotlib` publishes their nightlies, and then answer is that while I was working on https://github.com/scientific-python/nightly-wheels @jarrodmillman [had freshly](https://discuss.scientific-python.org/t/interest-in-github-action-for-scipy-wheels-nightly-uploads-and-removals/397/11) made https://anaconda.org/scientific-python-nightly-wheels/ and to do tests of the action workflow I used `matplotlib` as I already had that working well for `matplotlib` on the [`scipy-wheels-nightly` Anaconda org](https://anaconda.org/scipy-wheels-nightly).

@tacaswell @QuLogic I have a private IRIS-HEP Slack chat with @tacaswell from 2022-08-19 where I got permission to use `matplotlib` as a test upload (just wanted to make it very clear I wasn't going about distributing packages without express approval).

---

So this also means that until this feature is added

https://github.com/scientific-python/upload-nightly-action/blob/80af44feb2d73fe29869d46207c02f88885ea5b8/README.md?plain=1#L32

and

https://github.com/scientific-python/upload-nightly-action/blob/80af44feb2d73fe29869d46207c02f88885ea5b8/README.md?plain=1#L49-L54

are wrong (well, at least severely out of date) and `matplotlib` should not be used as an example.

If people are in agreement with me that this should be configurable (I would argue that unless you want this to be insular to **only** the Scientific Python org community this should not have a default but be a required argument) I'm happy to PR this.

betatim:
[...]

PR #12, opened 2023-05-31
CI: Add removal of old uploads

### LINKED ISSUE #4: Old wheels culling.
Another action that allows to upload to anaconda a is https://github.com/OpenAstronomy/publish-wheels-anaconda, in particular it allows to remove older wheels. 

I don't this we need to implement that as part of our action – too keep it  simple though as we are publishing everything on the same repo we could have a cron that cull all the old wheels from here.

matthewfeickert:
Resolves #4

Remove all but the last `N_LATEST_UPLOADS` (defaulted to `5`) package version uploads to the package index to ensure space. To do this, reply on the output form of `anaconda show` to be able to filter on the item delimiter character sequence for each version currently uploaded.

As an explicit example:

```console
$ anaconda show scientific-python-nightly-wheels/xarray
Using Anaconda API: https://api.anaconda.org
Name:    xarray
Summary: 
Access:  public
Package Types:  Standard Python
Versions:
   + 2023.5.1.dev2+gd8ec3a3f
   + 2023.5.1.dev5+gf96aca45
   + 2023.5.1.dev6+g69445c62
   + 2023.5.1.dev9+g609a9016
   + 2023.5.1.dev10+gf45eb733
   + 2023.5.1.dev11+gfcb81756
   + 2023.5.1.dev12+gb319d867
   + 2023.5.1.dev14+ge3db6164

To install this package with pypi run:
     pip install -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple xarray
```

shows that by filtering on `+`

```console
$ anaconda show scientific-python-nightly-wheels/xarray &> >(grep '+')
   + 2023.5.1.dev2+gd8ec3a3f
   + 2023.5.1.dev5+gf96aca45
   + 2023.5.1.dev6+g69445c62
   + 2023.5.1.dev9+g609a9016
   + 2023.5.1.dev10+gf45eb733
   + 2023.5.1.dev11+gfcb81756
   + 2023.5.1.dev12+gb319d867
   + 2023.5.1.dev14+ge3db6164
```

and then stripping `' + '`

```console
$ anaconda show scientific-python-nightly-wheels/xarray &> >(grep '+') | \
    awk '{print $2}'
2023.5.1.dev2+gd8ec3a3f
2023.5.1.dev5+gf96aca45
2023.5.1.dev6+g69445c62
2023.5.1.dev9+g609a9016
2023.5.1.dev10+gf45eb733
2023.5.1.dev11+gfcb81756
2023.5.1.dev12+gb319d867
2023.5.1.dev14+ge3db6164
```

one can obtain a newline separated list of all package uploads, where the _most recent_ uploads are listed last. After stripping off the `N_LATEST_UPLOADS` lines that correspond to the package versions to keep for testing

```console
$ anaconda show scientific-python-nightly-wheels/xarray &> >(grep '+') | \
    awk '{print $2}' | \
    head --lines -5 > remove-package-versions.txt
$ cat remove-package-versions.txt 
2023.5.1.dev2+gd8ec3a3f
2023.5.1.dev5+gf96aca45
2023.5.1.dev6+g69445c62
```

the remaining (older) package uploads can be removed with `anaconda remove`.

Most of this is taken from https://github.com/matplotlib/matplotlib/pull/23349 (c.f. https://github.com/scientific-python/upload-nightly-action/issues/4#issuecomment-1564749453).

To provide a local reproducible example of the behavior, `curl` the following attached Shell script into a `continuumio/miniconda3:latest` container with `anaconda-client` installed

<details>
<summary>example-pr-12.sh:</summary>

```bash
# example-pr-12.sh
#!/bin/bash

# fail on undefined variables
set -u
# Prevent pipe errors to be silenced
set -o pipefail
# Exit if any command exit as non-zero
set -e

# Input variables
ANACONDA_USER="scientific-python-nightly-wheels"
INPUT_ARTIFACTS_PATH="dist"
N_LATEST_UPLOADS=5

anaconda show "${ANACONDA_USER}" &> >(grep "${ANACONDA_USER}/") | \
    awk '{print $1}' | \
    sed 's|.*/||g' > package-names.txt

if [ -s package-names.txt ]; then
    while LANG=C IFS= read -r package_name ; do

        printf "\n# package: %s\n" "${package_name}"

        anaconda show "${ANACONDA_USER}/${package_name}" &> >(grep '+') | \
            awk '{print $2}' | \
            head --lines "-${N_LATEST_UPLOADS}" > remove-package-versions.txt

        if [ -s remove-package-versions.txt ]; then
            while LANG=C IFS= read -r package_version ; do
                echo "# Removing ${ANACONDA_USER}/${package_name}/${package_version}"
                # anaconda --token ${{ secrets.ANACONDA_TOKEN }} remove \
                #   --force \
                #   "${ANACONDA_USER}/${package_name}/${package_version}"
            done <remove-package-versions.txt
        fi

    done <package-names.txt
fi
```

```console
$ docker run --rm -ti continuumio/miniconda3:latest /bin/bash
(base) root@5d1ea993cbe9:/# conda install -y anaconda-client -c conda-forge
(base) root@5d1ea993cbe9:/# apt update && apt install -y curl
(base) root@5d1ea993cbe9:/# curl -sL https://github.com/scientific-python/upload-nightly-action/files/11618429/example-pr-12.sh.txt -o example-pr-12.sh
(base) root@5d1ea993cbe9:/# bash example-pr-12.sh 

# package: ipython

# package: matplotlib

# package: networkx

# package: numpy

# package: scikit-image

# package: scikit-learn

# package: scipy

# package: xarray

(base) root@5d1ea993cbe9:/# sed -i 's/N_LATEST_UPLOADS=5/N_LATEST_UPLOADS=2/g' example-pr-12.sh  # demo purposes
(base) root@5d1ea993cbe9:/# bash example-pr-12.sh

# package: ipython

# package: matplotlib
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3181+g9f17f3f851
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3182+gd6bad5f684

# package: networkx

# package: numpy
# Removing scientific-python-nightly-wheels/numpy/1.25.0.dev0+1532.g5d6c744db

# package: scikit-image

# package: scikit-learn

# package: scipy

# package: xarray
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev9+g609a9016
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev10+gf45eb733
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev11+gfcb81756
(base) root@5d1ea993cbe9:/# 
```

</details>

[example-pr-12.sh.txt](https://github.com/scientific-python/upload-nightly-action/files/11618429/example-pr-12.sh.txt)

I'm marking this as closing Issue #4 despite that it doesn't include a time cutoff as described in https://github.com/scientific-python/upload-nightly-action/issues/4#issuecomment-1569278784 given that I do not know of a `pip` or `anaconda-client` API that gets upload date information from the index (Probably exists somewhere though?) and this PR resolves all other matters as far as I understand it. We can revise or extend this if desired.

### DIFF
--- a/.github/workflows/remove-wheels.yml
+++ b/.github/workflows/remove-wheels.yml
@@ -1,0 +1,81 @@
+name: Remove old wheels
+
+on:
+  # Run daily at 1:23 UTC
+  schedule:
+  - cron:  '23 1 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# Needed for micromamba pickup
+defaults:
+  run:
+    shell: bash -l {0}
+
+env:
+  N_LATEST_UPLOADS: 5
+  ANACONDA_USER: "scientific-python-nightly-wheels"
+
+jobs:
+  remove:
+
+    runs-on: ubuntu-latest
+    # Set required workflow secrets in the environment for additional security
+    # https://github.com/scientific-python/upload-nightly-action/settings/environments
+    environment:
+      name: remove-old-wheels
+
+    steps:
+      - name: Install micromamba and anaconda-client
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-name: remove-wheels
+          create-args: >-
+            python=3.11
+            anaconda-client=1.11.2
+
+      - name: Show environment
+        run: env
+
+      - name: Show CLI API info
+        run: |
+          anaconda show --help
+          echo ""
+          anaconda remove --help
+
+      - name: Query package index for packages
+        run: |
+          anaconda show "${ANACONDA_USER}" &> >(grep "${ANACONDA_USER}/") | \
+              awk '{print $1}' | \
+              sed 's|.*/||g' > package-names.txt
+
+      - name: Remove old uploads to save space
+        run: |
+          # Remove all _but_ the last ${N_LATEST_UPLOADS} package versions
+          # N.B.: `anaconda show` places the newest packages at the bottom of the output
+          # of the 'Versions' section and package versions are preceded with a '   + '.
+
+          if [ -s package-names.txt ]; then
+              # Remember can't quote subshell as need to split on (space seperated) token
+              for package_name in $(cat package-names.txt); do
+
+                  echo -e "\n# package: ${package_name}"
+
+                  anaconda show "${ANACONDA_USER}/${package_name}" &> >(grep '+') | \
+                      awk '{print $2}' | \
+                      head --lines "-${N_LATEST_UPLOADS}" > remove-package-versions.txt
+
+                  if [ -s remove-package-versions.txt ]; then
+                      for package_version in $(cat remove-package-versions.txt); do
+                          echo "# Removing ${ANACONDA_USER}/${package_name}/${package_version}"
+                          anaconda --token ${{ secrets.ANACONDA_TOKEN }} remove \
+                            --force \
+                            "${ANACONDA_USER}/${package_name}/${package_version}"
+                      done
+                  fi
+
+              done
+          fi
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "github-actions"


action.yml
```diff
@@ -14,6 +14,10 @@ inputs:
   anaconda_nightly_upload_token:
     description: 'Token to upload to scientific python org'
     required: true
+  n_latest_uploads:
+    description: 'Number of latest uploads to keep on the package index'
+    required: false
+    default: 5
```
matthewfeickert, line 20: Defaulting to 5 as this was a what `matplotlib` agreed to in https://github.com/matplotlib/matplotlib/pull/23349.

entrypoint.sh
```diff
@@ -40,3 +40,33 @@ echo "Uploading wheels to anaconda.org..."
 
 anaconda -t $ANACONDA_TOKEN upload --force -u "$ANACONDA_ORG" "$INPUT_ARTIFACTS_PATH"/*.whl
 echo "Index: https://pypi.anaconda.org/$ANACONDA_ORG/simple"
+
+# Get the package name from the "${INPUT_ARTIFACTS_PATH}"/*.whl files
+if [ -d "${INPUT_ARTIFACTS_PATH}" ]; then
+    # Leave off *.whl in ls target to avoid directory name prefix on wheel_name
+    wheel_name=$(ls --almost-all "${INPUT_ARTIFACTS_PATH}" | head --lines 1)
```
matthewfeickert, line 47: I'm not 100% sure this is POSIX and not just Linux, but it works to not get the `dist/` prefix here

```console
$ ls --almost-all dist/ | head --lines 1
xarray-2023.5.1.dev10+gf45eb733-py3-none-any.whl
```

entrypoint.sh
```diff
@@ -40,3 +40,33 @@ echo "Uploading wheels to anaconda.org..."
...
+fi
+
+if [ ! -z "${INPUT_N_LATEST_UPLOADS}" ]; then
+  # Remove all _but_ the last "${INPUT_N_LATEST_UPLOADS}" package versions
+  # N.B.: `anaconda show` places the newest packages at the bottom of
+  # the output of the 'Versions' section and package versions are
+  # preceded with a '   + '.
+  anaconda show "${ANACONDA_ORG}/${package_name}" &> >(grep '+') | \
+      sed 's/.* + //' | \
+      head --lines "-${INPUT_N_LATEST_UPLOADS}" > remove-package-versions.txt
```
matthewfeickert, line 62: c.f. https://github.com/matplotlib/matplotlib/pull/23349 for a detailed breakdown of what is happening here.

entrypoint.sh
```diff
@@ -40,3 +40,33 @@ echo "Uploading wheels to anaconda.org..."
 
 anaconda -t $ANACONDA_TOKEN upload --force -u "$ANACONDA_ORG" "$INPUT_ARTIFACTS_PATH"/*.whl
 echo "Index: https://pypi.anaconda.org/$ANACONDA_ORG/simple"
+
+# Get the package name from the "${INPUT_ARTIFACTS_PATH}"/*.whl files
+if [ -d "${INPUT_ARTIFACTS_PATH}" ]; then
```
matthewfeickert, line 45: Opting for package name discovery over making it a required argument given https://github.com/scientific-python/upload-nightly-action/issues/8#issuecomment-1565022024 stressed the idea of simplicity of use for the end user.

tupui (commented):
Thank you @matthewfeickert. If that's the same approach that Matplotlib has been using successfully, then all good to me.

The only thing to me is that, IIUC this is to be done per project right? On SciPy and NumPy at least, we are not using actions to do the upload but we rely on CirrusCI. So we would need to add another action that would not upload and just check for removal? (In practice we overwrite our upload and only have a single version. But other packages might do differently.)

Should we have a nightly action on this repo that would run this for all packages?

matthewfeickert:
> The only thing to me is that, IIUC this is to be done per project right? On SciPy and NumPy at least, we are not using actions to do the upload but we rely on CirrusCI. So we would need to add another action that would not upload and just check for removal? (In practice we overwrite our upload and only have a single version. But other packages might do differently.)
> 
> Should we have a nightly action on this repo that would run this for all packages?

Ah I see that I have misunderstood what "centralise" meant in https://github.com/scientific-python/upload-nightly-action/issues/4#issuecomment-1569277569.

@tupui so for SciPy and NumPy you use CirrusCI for _all_ the uploads to https://anaconda.org/scientific-python-nightly-wheels? If so, then yes, this would all need to be a separate action.

I think that it would be better to do a 1-action-per-repo policy, so in that case this PR should get closed and I could port the relevant logic to the new repo (I'm not very good at naming things so `nightly-wheel-removal` or something of that sort).

tupui:
Let's continue the discussion on the issue before we make any decisions about closing etc. 😃

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,75 @@
...
+env:
+  N_LATEST_UPLOADS: 5
+  ANACONDA_USER: "scientific-python-nightly-wheels"
+
+jobs:
+  remove:
+
+    runs-on: ubuntu-latest
+    environment:
+      name: remove-old-wheels
```
matthewfeickert, line 27: Using a GHA environment to further restrict the possibility of accessing any API tokens.

Someone who has org admin permissions needs to create a GitHub Actions environment (https://github.com/scientific-python/upload-nightly-action/settings/environments) named "remove-old-wheels" and in there create an environment secret named `ANACONDA_TOKEN`. The value of `ANACONDA_TOKEN` can be set to the value of the `github-actions` Anaconda Cloud API token fund under https://anaconda.org/scientific-python-nightly-wheels/settings/access.

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,75 @@
...
+        package-name: [
+          "ipython",
+          "matplotlib",
+          "networkx",
+          "numpy",
+          "scikit-image",
+          "scikit-learn",
+          "scipy",
+          "xarray",
+        ]
```
matthewfeickert, line 39: These are the project that currently have wheels on https://anaconda.org/scientific-python-nightly-wheels.

bsipocz: I feel there should be a master list of the packages separately, rather than hard-wired in here (easier not missing to update when a new one is added)

matthewfeickert: @bsipocz If you do this then you will have to run sequentially instead of in parallel as I don't think(?) that it is possible to set GHA matrix values to variables. If I'm wrong that would be great. (Also feels a bit like once you need config files for a GHA workflow then this should live in a new repository).

matthewfeickert: You make a good point though that ideally we don't need to keep track of anything though and we just query the index itself for the project names.

bsipocz: Doing it with a query is great, but anything that is not hard-wired values in the workflow yaml is really an improvement.

tupui: Agreed, I don't think it's a concern to run this sequentially too. We can easily get the list with something like `anaconda show scientific-python-nightly-wheels`

stefanv:
Agreed that querying the index for the list of packages would make this a bit more robust against future package additions.

Aside, an alternative to `sed` magic:

```
| awk '{print $2}'
```

tupui (approved):
Thank you @matthewfeickert 🚀 That's good on my side. Approving as the suggestion is optional to me for now.

.github/workflows/remove-wheels.yml
matthewfeickert: Actually that was a simple over lunch fix, so good suggestion and yeah as @bsipocz pointed out this is actually faster (which in retrospect, if it wasn't there would be problems!).

tupui (approved):
Good on my side 🚀 thanks @matthewfeickert

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          anaconda show --help
+          echo ""
+          anaconda remove --help
+
+      - name: Query package index for packages
+        run: |
+          anaconda show "${ANACONDA_USER}" &> >(grep "${ANACONDA_USER}/") | \
+              awk '{print $1}' | \
+              sed 's|.*/||g' > package-names.txt
```
stefanv, line 51: Curious what the sed does here; would expect awk to print only the relevant part of what you need.

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Remove old uploads to save space
+        run: |
+          # Remove all _but_ the last ${N_LATEST_UPLOADS} package versions
+          # N.B.: `anaconda show` places the newest packages at the bottom of the output
+          # of the 'Versions' section and package versions are preceded with a '   + '.
+
+          if [ -s package-names.txt ]; then
+              while LANG=C IFS= read -r package_name ; do
```
stefanv, line 60: Alternative spelling that may be a bit easier on the eye:

```
for pkg in $(cat package-names.txt); do
  ... something with $pkg ...
done
```

.github/workflows/remove-wheels.yml
matthewfeickert: Yup, a comment is a good idea here. I'll add one.

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          environment-name: remove-wheels
+          create-args: >-
+            python=3.11
+            anaconda-client=1.11.2
+
+      - name: Show environment
+        run: env
+
+      - name: Show CLI API info
```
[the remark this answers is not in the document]
matthewfeickert: Not versions, but if someone is looking to the logs for debugging information it can be useful to understand what the `anaconda-client` CLI API is given that there documentation that exits for it (https://docs.anaconda.org/free/anacondaorg/commandreference/) is literally just the same thing. :/

.github/workflows/remove-wheels.yml
matthewfeickert: So indeed `awk` will given you the `USER/PACKAGE` information that is what is needed later

```console
$ anaconda show scientific-python-nightly-wheels &> >(grep "scientific-python-nightly-wheels/") | \
    awk '{print $1}'
scientific-python-nightly-wheels/ipython
scientific-python-nightly-wheels/matplotlib
scientific-python-nightly-wheels/networkx
scientific-python-nightly-wheels/numpy
scientific-python-nightly-wheels/scikit-image
scientific-python-nightly-wheels/scikit-learn
scientific-python-nightly-wheels/scipy
scientific-python-nightly-wheels/xarray
```

However, if you want to be able to refer to the package names by themselves without the user information then you would need to strip all of that off. This is what the use of `sed` does.

```console
$ anaconda show scientific-python-nightly-wheels &> >(grep "scientific-python-nightly-wheels/") | \
    awk '{print $1}' | \
    sed 's|.*/||g'
ipython
matplotlib
networkx
numpy
scikit-image
scikit-learn
scipy
xarray
```

There's other ways to do this of course, but as you've also noticed I reach for `sed` a lot out of habit/comfort.

.github/workflows/remove-wheels.yml
matthewfeickert: I'll fully admit that my use of the `while IFS= read -r line` pattern comes from blindly following [Shellcheck SC2013](https://www.shellcheck.net/wiki/SC2013) from years ago and at this point it is drilled into the fingers. I don't mind shifting things if it will help with the maintainability, but I think a nicer goal would be to eventually learn enough of the underlying `binstar_client` Python API from `anaconda-client` to rewrite things in not-a-shell language (assuming that somehow isn't worse for readability). :)

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Remove old uploads to save space
+        run: |
+          # Remove all _but_ the last ${N_LATEST_UPLOADS} package versions
+          # N.B.: `anaconda show` places the newest packages at the bottom of the output
+          # of the 'Versions' section and package versions are preceded with a '   + '.
+
+          if [ -s package-names.txt ]; then
+              while LANG=C IFS= read -r package_name ; do
+
+                  printf "\n# package: %s\n" "${package_name}"
```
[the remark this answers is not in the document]
matthewfeickert: The use of `printf` is my go-to for a 1-liner that allows for visual separation of print information to avoid things getting smushed together in the logs. E.g. the `\n` helps quite a bit IMO with output like

```
# package: ipython

# package: matplotlib
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3181+g9f17f3f851
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3182+gd6bad5f684

# package: networkx

# package: numpy
# Removing scientific-python-nightly-wheels/numpy/1.25.0.dev0+1532.g5d6c744db

# package: scikit-image

# package: scikit-learn

# package: scipy

# package: xarray
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev9+g609a9016
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev10+gf45eb733
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev11+gfcb81756
```

Though to make things less visually strange with switching back and forth between `echo` and `printf` I can unify things by just using `echo -e`. :+1:

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,80 @@
...
+  ANACONDA_USER: "scientific-python-nightly-wheels"
+
+jobs:
+  remove:
+
+    runs-on: ubuntu-latest
+    # Set required workflow secrets in the environment for additional security
+    # https://github.com/scientific-python/upload-nightly-action/settings/environments
+    environment:
+      name: remove-old-wheels
```
matthewfeickert, line 29: @stefanv as I don't have org level admin permissions can I get you to verify that the url in the comment (https://github.com/scientific-python/upload-nightly-action/settings/environments) correctly dumps you into the Environments list view and that someone has also setup the `remove-old-wheels` environment with the secret?

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
matthewfeickert, line 1: Oops! This is in the wrong place! One directory too far down.

.github/workflows/remove-wheels.yml
bsipocz: I agree with the comment above, it's so much easier to read the none "while" version.

.github/workflows/remove-wheels.yml
bsipocz: > also works, but as written (maybe there was a typo in the comment?)

I don't have anaconda, so there maybe multiple white spaces that the first awk works well on, but not the second. Either case I agree, more of us prefer awk than sed, and not because of performance 😄

matthewfeickert: >I don't have anaconda

c.f. the self contained example I wrote and included in the PR body that uses Docker.

bsipocz: sure, my point was that I don't have the anaconda-client to check on the whitespace in the command above. 

Bottom line, it should all be doable with one awk, but I suppose we bike shedded on it enough. Both Stefan and I made comments that made the shell more readable while using fewer tools and tricks, if you don't like the feedback, well, it's OK.

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Remove old uploads to save space
+        run: |
+          # Remove all _but_ the last ${N_LATEST_UPLOADS} package versions
+          # N.B.: `anaconda show` places the newest packages at the bottom of the output
+          # of the 'Versions' section and package versions are preceded with a '   + '.
+
+          if [ -s package-names.txt ]; then
+              # Remember can't quote subshell as need to split on (space seperated) token
+              for package_name in $(cat package-names.txt); do
```
matthewfeickert, line 63: @stefanv @bsipocz this uses the `cat` approach that you both found easier to read, but I've added this comment above it as if people quote subshells without thinking (which happens a lot in my experience) they will get a bug. Short example:

```console
$ echo -e "name1 abc\nname2" > /tmp/example.txt
$ cat /tmp/example.txt 
name1 abc
name2
$ while IFS= read -r line; do echo -e "line: ${line}"; done </tmp/example.txt  # desired behavior
line: name1 abc
line: name2
$ for line in "$(cat /tmp/example.txt)"; do echo -e "line: ${line}"; done  # undesired behavior
line: name1 abc
name2
$ for line in $(cat /tmp/example.txt); do echo -e "line: ${line}"; done  # potentially desired behavior
line: name1
line: abc
line: name2
```

This example adds in a factor of space separated tokens which doesn't appear in our particular use case to show the general example of how using `read` when you want _per line_ splitting helps protect you.

.github/workflows/remove-wheels.yml
matthewfeickert: @stefanv @jarrodmillman ping on this as you're the org owners (and so I think(?) are the only ones with the required level of permissions to do this).

matthewfeickert:
Thanks for the review @jarrodmillman! Does your approval mean that you were able to setup the `remove-old-wheels` GitHub Actions environment with the Anaconda cloud API token as an environment restricted secret?

.github/workflows/remove-wheels.yml
jarrodmillman: @matthewfeickert @stefanv @tupui I marked it as a draft to see if folks could take one last look (I don't want to accidentally want to delete stuff).

I followed the instructions below and I believe this ready to merge.

> Using a GHA environment to further restrict the possibility of accessing any API tokens.
> 
> Someone who has org admin permissions needs to create a GitHub Actions environment (https://github.com/scientific-python/upload-nightly-action/settings/environments) named "remove-old-wheels" and in there create an environment secret named `ANACONDA_TOKEN`. The value of `ANACONDA_TOKEN` can be set to the value of the `github-actions` Anaconda Cloud API token fund under https://anaconda.org/scientific-python-nightly-wheels/settings/access.

jarrodmillman:
See https://github.com/scientific-python/upload-nightly-action/pull/12#discussion_r1223294435

.github/workflows/remove-wheels.yml
matthewfeickert: > I followed the instructions below and I believe this ready to merge.

Great! Yeah, if the https://github.com/scientific-python/upload-nightly-action/settings/environments view looks similar to the following (from my fork) then this should be good to go.

env-view

stefanv (approved):

.github/workflows/remove-wheels.yml
```diff
@@ -0,0 +1,81 @@
+name: Remove old wheels
+
+on:
+  # Run daily at 1:23 UTC
+  schedule:
+  - cron:  '23 1 * * *'
+  workflow_dispatch:
```
stefanv, line 7: Thanks for adding this in: useful to be able to launch with the click of a button!

.github/workflows/remove-wheels.yml
jarrodmillman: It does!!

jarrodmillman: 2023-06-08T10:04:24,228096276-07:00

tupui: I am +1 here. Go for it I would say 😃

tupui (approved):

jarrodmillman:
Thanks @matthewfeickert !!!

matthewfeickert:
Just did a manual run with workflow dispatch and all looks good! 

```
# package: ipython

# package: matplotlib
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3181+g9f17f3f851
# Removing scientific-python-nightly-wheels/matplotlib/3.6.0.dev3182+gd6bad5f684
# Removing scientific-python-nightly-wheels/matplotlib/3.8.0.dev1157+g2f778fda79
# Removing scientific-python-nightly-wheels/matplotlib/3.8.0.dev1150+g3d3a1afd6d
Using Anaconda API: https://api.anaconda.org
# Removing scientific-python-nightly-wheels/matplotlib/3.8.0.dev1192+gbb335a115c
Using Anaconda API: https://api.anaconda.org
# Removing scientific-python-nightly-wheels/matplotlib/3.8.0.dev1211+gdcb8180edc
Using Anaconda API: https://api.anaconda.org
# Removing scientific-python-nightly-wheels/matplotlib/3.8.0.dev1228+g19d93b7876
Using Anaconda API: https://api.anaconda.org

# package: networkx

# package: numpy
# Removing scientific-python-nightly-wheels/numpy/1.25.0.dev0+1532.g5d6c7[44](https://github.com/scientific-python/upload-nightly-action/actions/runs/5224131691/jobs/9431935913#step:6:45)db
Using Anaconda API: https://api.anaconda.org

# package: pandas

# package: scikit-image

# package: scikit-learn

# package: scipy

# package: statsmodels

# package: xarray
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev9+g609a9016
Using Anaconda API: https://api.anaconda.org
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev10+gf[45](https://github.com/scientific-python/upload-nightly-action/actions/runs/5224131691/jobs/9431935913#step:6:46)eb733
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev11+gfcb81756
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev12+gb319d867
# Removing scientific-python-nightly-wheels/xarray/2023.5.1.dev14+ge3db6164
```

Should be good to go for nightly cron jobs now and then just need to make sure that the workflow doesn't get shut off due to lack of repo activity.

bsipocz:
It could also push a timestamp of its last run, maybe in a separate, empty branch, or gh-pages?

matthewfeickert:
> It could also push a timestamp of its last run, maybe in a separate, empty branch, or gh-pages?

Yup, but then you're accumulating a daily commit history on that other branch that still lives in the repo for no real reason. Is it easy to ignore? Sure, but seems kinda unnecessary. As you point out though, you could use `gh-pages` as a way to throw this stuff into the void in some sense, but if for any reason in the future you want to use the `gh-pages` for an actual website then you need to shift things around.

Personally, I would put this problem in the "kick it down the road until it becomes a problem" bin, as for the multiple projects that I maintain that are are at this stage LTS Docker images this hasn't come up much. Though if anyone has been nerdsniped by this conversation please go for it! :wink:

bsipocz:
I had repos where we run into this in practice, and wondered why the actions stopped working. So I would just like to avoid being on that road again :)

stefanv:
One commit every 50 days on a `keepalive` branch isn't too bad :)

https://github.com/marketplace/actions/keepalive-workflow

bsipocz:
Ahh, nice, thanks for the link. I was starting to hack something together, but it's probably much easier to just do the action. I'll open a PR for this.

PR #13, opened 2023-05-31
DOC: Advocate for using action from tagged release commit shas

### LINKED ISSUE #7: Consider not advocating using action from `main`
At the moment the `v0.1.0` docs advocate that users use the action from `main` and not from a specific tag.

https://github.com/scientific-python/upload-nightly-action/blob/80af44feb2d73fe29869d46207c02f88885ea5b8/README.md?plain=1#L10-L17

This seems like a security issue that should be avoided and that @webknjaz has discussed before for https://github.com/pypa/gh-action-pypi-publish.

A more secure option is to advocate that users use release tags and dependabot to check for updates to the GitHub Actions that they use and then consider from the PR if the update in the tag is something they are ready to accept. Example `.github/dependabot.yml` (taken from [Scikit-HEP developer recommendations](https://scikit-hep.org/developer/gha_basic#updating) (thanks @henryiii!)):

```yaml
version: 2
updates:
  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

matthewfeickert:
Resolves #7 

* For security best practices, advocate that users of the action use it from known commit shas that correspond to tagged releases.
* Advocate that users use a Dependabot config file to update the action on new tags. This will bump the commit sha and also bump the release tag in the comment of the commit sha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,10 @@
 # Nightly upload
 
-This provides a standard GitHub Action to upload nightly builds to the
-scientific-python nightly channel.
+This is a GitHub Action that uploads nightly builds to the [scientific-python nightly channel][],
+as recommended in [SPEC4 — Using and Creating Nightly Wheels][].
 
-In your Continuous Integration pipeline once you've built you wheel, you can
-use the following snippet to upload to our central nightly repository:
-
-<!-- c.f. https://github.com/scientific-python/upload-nightly-action/pull/13 and
-https://github.com/matplotlib/matplotlib/pull/26023#discussion_r1212539700
-for short summary of why using commit SHA -->
+In a GitHub Actions workflow (`.github/workflows/*.yaml`), use the
+following snippet to upload built wheels to the repository:
 
 ```yml
 jobs:
@@ -21,8 +17,14 @@ jobs:
         anaconda_nightly_upload_token: ${{secrets.UPLOAD_TOKEN}}
 ```
 
-It is [recommended that Dependabot is used][] to keep the GitHub Action updated
-to the latest release by using a `.github/dependabot.yml` config file similar to
+Note that we recommend pinning the action against a specific SHA
+(rather than a tag), to guard against the unlikely event of upstream
+being compromised.
+
+# Updating the action
+
+You can [use Dependabot to keep the GitHub Action up to date][],
+with a `.github/dependabot.yml` config file similar to:
 
 ```yaml
 version: 2
@@ -34,16 +36,18 @@ updates:
       interval: "weekly"
 ```
 
-To request access to the repository please open an issue on [this action
+# Access
+
+To request access to the repository, please open an issue on [this action's
 repository](https://github.com/scientific-python/upload-nightly-action). You can
 then generate a token at `https://anaconda.org/scientific-python-nightly-wheels/settings/access`
-with _Allow write access to the API site_ and _Allow uploads to Standard Python repositories_
-permissions and add the token as a secret to your GitHub repository.
+with permissions to _Allow write access to the API site_ and _Allow uploads to Standard Python repositories_,
+and add the token as a secret to your GitHub repository.
 
 # Using nightly builds in CI
 
-To test those nightly build, you can use the following command to install from
-the nightly package.
+To test against nightly builds, you can use the following command to install from
+the nightly repository:
 
 ```sh
 python -m pip install \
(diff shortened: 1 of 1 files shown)


stefanv (approved):

README.md
```diff
@@ -11,12 +11,25 @@ jobs:
   steps:
     ...
     - name: Upload wheel
-      uses: scientific-python/upload-nightly-action@main
+      uses: scientific-python/upload-nightly-action@8f0394fd2aa0c85d7364a9958652e8994e06b23c # 0.1.0
```
bsipocz, line 14: could this be a major version instead, the same we do for checkout, or python, or etc?

README.md
```diff
@@ -11,12 +11,25 @@ jobs:
...
       with:
         artifacts_path: dist
         anaconda_nightly_upload_token: ${{secrets.UPLOAD_TOKEN}}
 ```
 
+It is recommended that [Dependabot][] is used to keep the GitHub Action updated
+to the latest release by using a `.github/dependabot.yml` config file similar to
+
+```yaml
```
bsipocz, line 23: can we link to a docs instead keeping this example here? E.g. I wonder about maintenance when e.g. `version: 2` becomes outdated, etc.

bsipocz (commented):
The notion sounds fine to me, and we talked about doing it anyway while writing the action, but first just wanted to get something out of the door asap.

README.md
stefanv: It depends on how much you trust the project; releases / tags can be removed and replaced, while SHAs are a bit harder to fake. But if you trust official releases made by the org, and you just want to review the new action to make sure nothing big changed, versions are fine.

bsipocz: well, I would rather like to think that the releases made in the `scientific-python` org are official and trustworthy :)

bsipocz: (after all, we all use a lot of `main` actions, e.g. the artifact upload one, which is I agree is not super ideal as a best practice)

stefanv: In that case, if you trust that bunch, versions seem fine ;)

matthewfeickert: I think that @ksunden gave a good and concise summary in https://github.com/matplotlib/matplotlib/pull/26023#discussion_r1212539700. This is not about trusting orgs or not trusting orgs. It is about trying to be extra security cautious for anything that is publishing distributions to any package index. Is this overkill? Yeah, for sure in my mind, but it also seems like a pattern with no downsides other than you have to have your eyes track a little further across the screen to read the associated tag.

> after all, we all use a lot of main actions, e.g. the artifact upload one, which is I agree is not super ideal as a best practice

I will fully agree with you though that tags are better than `main`, and patch version tags are better than major version tags. Though I'm also willing to agree that there should be additional security hardening than using a `latest` tag Docker image

https://github.com/scientific-python/upload-nightly-action/blob/dbd5a4da48dfa98ef9f67c571248ea4891c651fd/Dockerfile#L1

and so using the SHAs only really helps later releases (`0.1.0` is the first commit in the repo).

README.md
matthewfeickert: > can we link to a docs instead keeping this example here?

Sure! @henryiii just migrated https://scikit-hep.org/developer/gha_basic#updating to https://learn.scientific-python.org/development/guides/gha_basic/#updating so we can use the shiny new Scientific Python docs for this. :+1:

Though I would both link it _and_ keep it in the README as it seemed from Issue #8 that making things easy to copy and paste for adoption was something that while not an explicit goal would be the recommendation.

> I wonder about maintenance when e.g. version: 2 becomes outdated, etc.

I'll let @henryiii correct me but I believe this is a very slow release cycle so I think the answer is that the maintenance burden is minimal.

README.md
bsipocz: Thanks for the detailed reply. To make this bikeshed really shiny, could I ask you to include some of Kyle's concise commentary as a comment in this readme? Or even just the link you have above would be useful.

README.md
bsipocz: OK, let's remove it if/when it becomes an issue.

bsipocz (approved):

README.md
```diff
@@ -3,20 +3,37 @@
...
 scientific-python nightly channel.
 
-In your Continuous Intregration pipeline once you've built you wheel, you can
+In your Continuous Integration pipeline once you've built you wheel, you can
 use the following snippet to upload to our central nightly repository:
 
+<!-- c.f. https://github.com/scientific-python/upload-nightly-action/pull/13 and
+https://github.com/matplotlib/matplotlib/pull/26023#discussion_r1212539700
+for short summary of why using commit SHA -->
+
```
matthewfeickert, line 12: @bsipocz ...and after a long delay given too many meetings and weekend travel this is in now. Look good?

jarrodmillman:
Feel free to follow-up on this in a future PR, but this looks like it is already a good improvement. Thanks!!

PR #16, opened 2023-06-08
In `environment.yaml`, add pip as dependency before configuring flags

stefanv:
The previous approach works with mamba and conda, but not with micromamba. So this adds `pip` back as an explicit dependency, before configuring pip installation flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ index in your `environment.yml`:
 ```yml
 name: test
 dependencies:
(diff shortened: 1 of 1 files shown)


jarrodmillman (approved):

stefanv:
Ref: https://github.com/mamba-org/mamba/issues/2574#issuecomment-1583115007

PR #18, opened 2023-06-09
CI: Add workflow to avoid scheduled workflows becoming disabled

### LINKED ISSUE #17: Avoid scheduled workflows becoming disabled due to lack of repo activity
> I didn't realize inactive repos have their actions switched off. What do we need to do; a monthly empty commit?

_Originally posted by @stefanv in https://github.com/scientific-python/upload-nightly-action/issues/12#issuecomment-1584931283_

matthewfeickert:
Resolves #17 

Run a cron job on a monthly basis that force pushes a commit to a `keep-alive` branch.

(There might be more interesting or intelligent ways to do this, but this works. Revisions welcome!)

### DIFF
--- a/.github/workflows/keep-alive.yml
+++ b/.github/workflows/keep-alive.yml
@@ -1,0 +1,38 @@
+name: Keep workflows alive
+
+on:
+  # Run on the first of each month at 1:23 UTC
+  schedule:
+  - cron:  '23 1 1 * *'
+  workflow_dispatch:
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Create a minimal payload
+      run: |
+        mkdir _pass
+        touch _pass/pass.txt
+
+    - name: Keep workflow alive
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_branch: keep-alive
+        publish_dir: _pass
+        enable_jekyll: true  # avoid extra files
+        force_orphan: true  # overwirte
+        user_name: 'github-actions[bot]'
+        user_email: 'github-actions[bot]@users.noreply.github.com'
+        commit_message: Keep repo GitHub Actions alive


.github/workflows/keep-alive.yml
```diff
@@ -0,0 +1,36 @@
...
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Create a minimal payload
+      run: |
+        mkdir _pass
+        touch _pass/pass.txt
```
matthewfeickert, line 26: `peaceiris/actions-gh-pages` requires a non-empty payload directory, so this just provides  payload of an empty file.

.github/workflows/keep-alive.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: Keep workflow alive
+      uses: peaceiris/actions-gh-pages@v3
```
matthewfeickert, line 29: I think the security risk of using a `@v3` tag here is probably fine as we aren't deploying thing, but as we _are_ pushing back to the repo I can see some motivation to go with a more security cautious approach. I'll defer to others here.

.github/workflows/keep-alive.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_branch: keep-alive
+        publish_dir: _pass
+        enable_jekyll: true  # avoid extra files
+        force_orphan: true  # overwirte
```
matthewfeickert, line 35: force pushes to the branch

.github/workflows/keep-alive.yml
```diff
@@ -0,0 +1,38 @@
...
+        user_name: 'github-actions[bot]'
+        user_email: 'github-actions[bot]@users.noreply.github.com'
```
matthewfeickert, line 37: Avoids the commit being associated with a real user (this really doesn't matter much IMO, but some people like it to make it clear that "hey this wasn't me this was an automated system").

.github/workflows/keep-alive.yml
stefanv: I don't see why this is less risky than other cases: the action gets your secret and can push a new tag to do anything with it.

stefanv (approved):
This looks good, thanks!

Any reason not to use the official GH pages action?

.github/workflows/keep-alive.yml
Carreau: I think what @matthewfeickert is `@v3` pins to any v3, while `@v3.9.1` or `@<commit-ish>` would pin to a given version so if upstream is compromised we are safe.

stefanv: Presumably v3.9.1 is a tag, so only hashes are secure in this sense?

matthewfeickert: > the action gets your secret and can push a new tag to do anything with it.

The [default `GITHUB_TOKEN` secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) created by GitHub Actions at runtime has its permissions limited to the repository that contains the workflow. So it can't perform arbitrary actions. Though, it can write (and so rewrite) commits, so given that

> so only hashes are secure in this sense?

yes, probably.

matthewfeickert:
> Any reason not to use the official GH pages action?

@stefanv If you mean https://github.com/actions/deploy-pages, that GitHub Action doesn't allow for configuration of the target branch. It will only deploy to the `gh-pages` branch, but if we want the ability to reserve that for later use we need to target a different branch which `peaceiris/actions-gh-pages` allows.

stefanv:
That makes sense, thanks!

PR #19, opened 2023-06-13
MNT: Use commit hash and run only on main repository

matthewfeickert:
In follow up to https://github.com/scientific-python/upload-nightly-action/pull/18/files#r1225142853, use the commit hash for `peaceiris/actions-gh-pages` `v3.9.3`. To avoid nightly failures on the `.github/workflows/remove-wheels.yml` on forks, which lack access to

https://github.com/scientific-python/upload-nightly-action/blob/b1c5a48e464ab8a45a1899cfaefc21548923ee6d/.github/workflows/remove-wheels.yml#L74

restrict the workflow to run only on the main repository by requiring the owner be the `scientific-python` GitHub org.

### DIFF
--- a/.github/workflows/keep-alive.yml
+++ b/.github/workflows/keep-alive.yml
@@ -26,7 +26,7 @@ jobs:
         touch _pass/pass.txt
 
     - name: Keep workflow alive
-      uses: peaceiris/actions-gh-pages@v3
+      uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 #v3.9.3
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_branch: keep-alive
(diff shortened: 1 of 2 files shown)


.github/workflows/keep-alive.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         touch _pass/pass.txt
 
     - name: Keep workflow alive
-      uses: peaceiris/actions-gh-pages@v3
+      uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 #v3.9.3
```
matthewfeickert, line 29: This will get updated by `.github/dependabot.yml` along with the rest of the GitHub Actions used in these workflows.

jarrodmillman (approved):

PR #24, opened 2023-07-18
DOC: Fix broken url to Scientific Python Development Guide

matthewfeickert:
* At some point urls were changed to use '-' over '_'.


matthewfeickert:
I'm going to approve and merge this myself given it is the only link in the README that is broken. As always, PRs approved by a single core dev can be reverted as needed by the rest of the team.

Issue #15, opened 2023-06-02
Add tests and CI


matthewfeickert:
> Also this GHA has no CI at all yet. :grimacing: 

_Originally posted by @matthewfeickert in https://github.com/scientific-python/upload-nightly-action/pull/13#discussion_r1213785627_

CI and tests are always good to have, especially when publishing distributions.

I had a selection of tests for the old action that I had made, so I can recreate those here. Perhaps there can also be a fake package namespace on the index for testing uploads(?).

PR #26, opened 2023-07-19
FIX: Ensure USER has permissions for chmod in Dockerfile

### LINKED ISSUE #25: Fix permissions given default user is non-root
The default user in 

https://github.com/scientific-python/upload-nightly-action/blob/e09d83f8a9d79b1998dd94cd9fd5115162961820/Dockerfile#L1

is for security reasons a non-root user with `uid` `1000`:

```console
$ docker run --rm -ti mambaorg/micromamba:1.4.9-bullseye-slim
(base) mambauser@b82a37bc6fff:/tmp$ id
uid=1000(mambauser) gid=1000(mambauser) groups=1000(mambauser)
(base) mambauser@b82a37bc6fff:/tmp$ 
```

I forgot this in PR #23 and failed to implement Issue #15, which is causing errors for people who are working at `@main` and not secure pins (sorry `networkx` and `xarray`).

I will need to fix this tomorrow to ensure that things are running in login shells and with proper permissions.

matthewfeickert:
Resolves #25 

* For security reasons the default user in `mambaorg/micromamba:1.4.9-bullseye-slim` is non-root with uid 1000 (`mambauser`) but this requires the user permissions to be escalated to one with `chmod` powers. The simplest way to do this is to `chown` to `mambauser` on `COPY`.

* To ensure that the `micromamba` environment is setup correctly on entry fallback to the `mambaorg/micromamba`'s `ENTRYPOINT` and set the `/entrypoint.sh` as `CMD`.
   - c.f. https://github.com/mamba-org/micromamba-docker/blob/675aa44d37707d67b1c3f5ef899a3cc103c73e40/_entrypoint.sh


matthewfeickert:
Tests are needed in general (c.f. Issue #15), but if people would like this in sooner to fix things then a simple local test is

```console
$ docker build -f Dockerfile -t scientific-python/upload-nightly-action:debug .
$ docker run --rm -ti --env INPUT_ANACONDA_NIGHTLY_UPLOAD_TOKEN="fail" --env INPUT_ARTIFACTS_PATH="." scientific-python/upload-nightly-action:debug
```

which will run all the way through until it correctly errors on

https://github.com/scientific-python/upload-nightly-action/blob/e09d83f8a9d79b1998dd94cd9fd5115162961820/entrypoint.sh#L46-L49

for not being able to upload.

Dockerfile
```diff
@@ -1,12 +1,15 @@
...
 # Use C.UTF-8 locale to avoid issues with unicode encoding
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-COPY entrypoint.sh /entrypoint.sh
+COPY --chown=mambauser entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT [ "/usr/local/bin/_entrypoint.sh" ]
```
stefanv, line 14: I read your PR description, but am still fuzzy on what is happening in these two lines. Maybe a comment would help?

Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+COPY --chown=mambauser entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
-ENTRYPOINT ["/entrypoint.sh"]
+# The mambaorg/micromamba base image's entrypoint is
+# /usr/local/bin/_entrypoint.sh which ensures the shell environment is
+# correctly set for micromamba to be accessible by the given user.
+# c.f. https://github.com/mamba-org/micromamba-docker/blob/604ebafb09543a3d852e437886f1c782f0367911/_entrypoint.sh
+# Instead of replicating this, continue to use it as the ENTRYPOINT
+# and then pass the action's entrypoint as CMD.
```
matthewfeickert, line 19: @stefanv Let me know if this is helpful or still opaque. I'm happy to revise this!

Dockerfile
stefanv: I think the confusion is calling ours entrypoint as well; it really is the command we execute AFTER entrypoint, so we can just give it a logical name that describes what it does?

But, this is already way better, so we could merge this, perhaps with a note to explain the distinction between ENTRYPOINT and CMD.

I.e. we use entrypoint to set env variables, and the cmd does whatever action we wish to execute.

stefanv (approved):

jarrodmillman (approved):

stefanv:
Thanks, @matthewfeickert!

PR #27, opened 2023-07-19
CI: Add test workflow for action

### LINKED ISSUE #15: Add tests and CI

matthewfeickert:
Resolves #15 

* Add simplest example package to tests so that it can be built for upload to Anaconda cloud as a test of the action.
   - Use `setuptools` as the build backend over something like `hatchling` given the simple nature of the `test-package` given that `anaconda-client` `v1.12.0` doesn't understand valid metadata from PEP 518 build backends other than `setuptools`. c.f. https://github.com/Anaconda-Platform/anaconda-client/issues/676

* Add CI GitHub Action workflow to test the functionality of the action as shown in the README.

https://github.com/scientific-python/upload-nightly-action/blob/4792bc241c0a689e41f0fc4bf0953f26b8324c7b/README.md?plain=1#L10-L17

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,78 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  # Run weekly at 1:23 UTC
+  schedule:
+  - cron:  '23 1 * * 0'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: "test upload via action"
+    runs-on: ubuntu-latest
+    if: github.repository == 'scientific-python/upload-nightly-action'
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+
+      - name: Install python-build and twine
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install build twine
+          python -m pip list
+
+      - name: Build a wheel and a sdist
+        run: |
+          PYTHONWARNINGS=error,default::DeprecationWarning python -m build --outdir ./dist tests/test_package
+
+      - name: Verify the distribution
+        run: twine check --strict dist/*
+
+      - name: List contents of sdist
+        run: python -m tarfile --list dist/test-package-*.tar.gz
+
+      - name: List contents of wheel
+        run: python -m zipfile --list dist/test_package-*.whl
+
+      - name: Test upload
+        uses: ./
+        with:
+          artifacts_path: dist
+          anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
+
+  cleanup:
+    runs-on: ubuntu-latest
+    needs: [test]
+    # Set required workflow secrets in the environment for additional security
+    # https://github.com/scientific-python/upload-nightly-action/settings/environments
+    environment:
+      name: remove-old-wheels
+
+    steps:
+      - name: Install micromamba and anaconda-client
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-name: remove-wheels
+          create-args: >-
+            python=3.11
+            anaconda-client
+
+      - name: Remove test package upload
+        shell: bash -l {0}
+        run: |
+            anaconda --token ${{ secrets.ANACONDA_TOKEN }} remove \
+              --force \
+              "scientific-python-nightly-wheels/test-package"
--- a/tests/test_package/README.md
+++ b/tests/test_package/README.md
@@ -1,0 +1,1 @@
+# Test README
--- a/tests/test_package/pyproject.toml
+++ b/tests/test_package/pyproject.toml
@@ -1,0 +1,13 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "test-package"
+version = "0.0.1"
+authors = [
+  { name="Scientifc Python Developers", email="null@example.com" },
+]
+description = "Test project for testing GitHub Action"
+readme = "README.md"
+requires-python = ">=3.8"


.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: python -m tarfile --list dist/debug-*.tar.gz
+
+      - name: List contents of wheel
+        run: python -m zipfile --list dist/debug-*.whl
+
+      - name: Test upload
+        uses: ./
+        with:
+          artifacts_path: dist
+          anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
```
matthewfeickert, line 54: Currently broken by https://github.com/Anaconda-Platform/anaconda-client/issues/676

matthewfeickert, line 54: @stefanv @jarrodmillman As only org owners have access to the GitHub Actions secrets, can one of you please go to https://github.com/scientific-python/upload-nightly-action/settings/secrets/actions and add a new secret named `UPLOAD_TOKEN` that has the value of the Anaconda Clout API token named `github-actions` on https://anaconda.org/scientific-python-nightly-wheels/settings/access ?

tests/debug_project/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "debug"
```
matthewfeickert, line 6: I'm overall not the best at naming things, so I welcome suggestions of what this test package should be called so that people understand that it exists only for testing and should ignore it. Though `test_package_ignore` could work.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
+name: CI
+
+on:
+  push:
+    # branches: [ main ]
```
matthewfeickert, line 5: Using for debugging. Make sure to revert comment before merge.

.github/workflows/ci.yml
matthewfeickert: Resolved (ish) in https://github.com/Anaconda-Platform/anaconda-client/issues/676#issuecomment-1644930931.

matthewfeickert: Gentle ping on this, as it is needed for testing.

jarrodmillman: Sorry, looking into it now.

matthewfeickert: Nothing to be sorry about!

jarrodmillman: Done.

matthewfeickert: @jarrodmillman Hm. We can follow up on this tomorrow or next week, but for some reason the `UPLOAD_TOKEN` secret you set isn't getting picked up as

https://github.com/scientific-python/upload-nightly-action/blob/a2ba178f650e68bd6f9d8e2b7d59d16ffd257c29/cmd.sh#L18-L25

is erroring out in the CI with

```
...
+ echo 'Getting anaconda token from github secrets...'
+ ANACONDA_ORG=scientific-python-nightly-wheels
Getting anaconda token from github secrets...
+ ANACONDA_TOKEN=
+ '[' -z '' ']'
+ echo 'ANACONDA_TOKEN is empty , exiting...'
ANACONDA_TOKEN is empty , exiting...
+ exit -1
```

jarrodmillman: 2023-07-20T21:44:12,772309624-07:00

The `ANACONDA_TOKEN` environmental secret already existed.

I added the secret as `UPLOAD_TOKEN`  repository secret.

matthewfeickert: > I added the secret as `UPLOAD_TOKEN` repository secret.

Yup, that was the right thing to do. Hm, strange. I'll think on this tomorrow.

jarrodmillman: I just readded the secret to make sure I put in the correct one.

matthewfeickert: @jarrodmillman ah no sorry I wasn't thinking. You did everything right, but this PR is coming from my **fork** and so as there isn't already a workflow to run locally this is using my fork's secret credentials which are missing this information. I just verified this by pushing a `debug/is-this-repo-view-related` branch to this repo and it sees the secret correctly (:+1:).

However, this is still failing given https://github.com/Anaconda-Platform/anaconda-client/issues/676 with

```
...
Uploading wheels to anaconda.org...
Using "scientific-python-nightly-wheels" as upload username
Processing "dist/test_package-0.0.1-py3-none-any.whl"
Detecting file type...
File type is "Standard Python"
Extracting standard python attributes for upload
Error:  Trouble reading metadata from "dist/test_package-0.0.1-py3-none-any.whl". Is this a valid standard python package?
Error:  Trouble reading metadata from "dist/test_package-0.0.1-py3-none-any.whl". Is this a valid standard python package?
```

so I'll need Anaconda to clarify what the minimum metadata they need is. I'll keep this in draft until I get that verified.

matthewfeickert: Unfortunately, `anaconda-client` doesn't know how to deal with valid metadata from non-setuptools PEP 518 backends. :(

https://github.com/Anaconda-Platform/anaconda-client/issues/676#issuecomment-1645011255

So this forces us to use `setuptools` for the time being, which is annoying.

matthewfeickert:
Note to reviewers: The CI fails because GitHub Actions is using my fork's credentials to try to run this (as this workflow doesn't exist on the main repo yet). I pushed this branch under a different name as a test to the main repo and it was able to run and upload `test-package` correctly: https://anaconda.org/scientific-python-nightly-wheels/test-package

So this is working as I would expect, and so is ready for your reviews. :+1:

matthewfeickert:
Gentle ping for review when people have time.

MridulS (approved):
LGTM!

Maybe we should link this test action from the root README file too?
For example for networkx, we just build and upload. Maybe we should add the verification and debugging steps too 😅

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  # Run daily at 1:23 UTC
+  schedule:
+  - cron:  '23 1 * * *'
```
bsipocz, line 10: Does this really have to run daily? Weekly, or biweekly should be enough for such a repo.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: python -m tarfile --list dist/test-package-*.tar.gz
+
+      - name: List contents of wheel
+        run: python -m zipfile --list dist/test_package-*.whl
+
+      - name: Test upload
+        uses: ./
+        with:
+          artifacts_path: dist
+          anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
```
bsipocz, line 54: So where would this test package end up being uploaded? Is there any cleanup planned once the upload is successful?

matthewfeickert:
> Maybe we should link this test action from the root README file too? For example for networkx, we just build and upload. Maybe we should add the verification and debugging steps too

@MridulS We could do that, but I would suggest not using this GHA workflow but creating a template that anyone in the org could use, as _this_ workflow is focused on just testing the test package. There are already other projects and Actions that do this and more, like https://github.com/hynek/build-and-inspect-python-package, but I could understand the interest in an action template that would be more specific to a minimum set of verification stages that people can then extend if they want. This would be a separate project though IMO.

While I have personally found benefits in `pyhf` from having these verification steps I don't think that anyone _not_ using them now should feel bad. Though I agree if we can make adoption even easier that's great.

.github/workflows/ci.yml
matthewfeickert: Why not?

.github/workflows/ci.yml
matthewfeickert: > So where would this test package end up being uploaded?

c.f. https://github.com/scientific-python/upload-nightly-action/pull/27#issuecomment-1645021980

> Is there any cleanup planned once the upload is successful?

No, but does there need to be?

bsipocz: Yes. I would definitely like to avoid polluting the user deployed space with testing artifacts.

.github/workflows/ci.yml
bsipocz: Because we don't expect fast pace breakings here, as we practically have very few dependencies, and it's totally unnecessary use of resource.

In fact, as I see all versions are pinned, so then there is zero sense in running the cron.

matthewfeickert: > In fact, as I see all versions are pinned, so then there is zero sense in running the cron.

No, the action isn't fully specified (that requires a lock file, which is an upcoming PR), and the `ci` workflow uses unpinned versions of other actions that you want to know about. So testing on CRON is a good idea. The resource use is trivial IMO, but if it is really offensive we can move it to weekly.

.github/workflows/ci.yml
bsipocz: I'm not suggesting another location, but to do a cleanup. Easy and simple and the pollution will be only present when there is a problem.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  # Run weekly at 1:23 UTC
+  schedule:
+  - cron:  '23 1 * * 7'
```
matthewfeickert, line 10: CRON is now weekly.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+          environment-name: remove-wheels
+          create-args: >-
+            python=3.11
+            anaconda-client
+
+      - name: Remove test package upload
+        run: |
+            anaconda --token ${{ secrets.ANACONDA_TOKEN }} remove \
+              --force \
+              "scientific-python-nightly-wheels/test-package"
```
matthewfeickert, line 77: This should now remove the `test-package` package that gets uploaded just previously in the `test` job.

.github/workflows/ci.yml
matthewfeickert: Oh this is apparently non-standard CRON syntax? Let me fix this. :(

matthewfeickert: `'23 1 * * 7'` is nonstandard because of the `7` so I swapped this to `0`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install micromamba and anaconda-client
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-name: remove-wheels
+          create-args: >-
+            python=3.11
+            anaconda-client
+
+      - name: Remove test package upload
+        shell: bash -l {0}
```
matthewfeickert, line 74: Okay final burst of noise (sorry) to add in this `shell` line so that `conda` works in Actions. This works as expected (just tested on) a debug branch on the main project.

bsipocz:
> if you get time before the end of the week can you review if the changes I made given your original comments address them?

it would be super nice to allow a quick re-review if commits made after the review were not squashed into the original one. We do squash on merge anyway, so squashing in the PR has no actual benefits.

matthewfeickert:
> it would be super nice to allow a quick re-review if commits made after the review were not squashed into the original one.

@bsipocz see https://github.com/scientific-python/upload-nightly-action/pull/27#pullrequestreview-1557874446

bsipocz (approved):
OK, I can't spot any more issues, let's try and see it indeed passes after merge.

PR #32, opened 2023-08-11
MNT: Use conda-lock lock file for reproducible environment

matthewfeickert:
This PR does 3 things:

1. Define the uploader's high level dependencies in a `conda` `environment.yml` file.
2. Provide a Bash script (could have also been a `noxfile` but I thought that people might want to keep things less fancy and just go with something very straightforward) that uses the **same** Docker image that the Action's Dockerfile uses as a base image to install `conda-lock` and then create a `conda-lock.yml` lock file for the `environment.yml` which fully specifies the environment down to the hash level.
3. Using this lock file in the Action's Dockerfile have the action install the specified locked environment efficiently and reproducibly.

## Advantages of this method

* Updating to a new release of `anaconda-client` can be tested through the CI tests with high confidence that no breaking changes will follow. `anaconda-client` has [many dependencies](https://github.com/Anaconda-Platform/anaconda-client/blob/45fb0a363ba7833deccee6db82a26a0b51a7ca75/requirements.txt), that are (correctly) not restricted, which means that specifying the `anaconda-client` version number alone is not enough as all these dependencies can float beneath it and in addition to potentially breaking any `anaconda-client` release in the future can be a security risk. The only way around this is through a hash level lock file which specifies the exact binary that should be installed.
   - With no blame towards the Anaconda open source team, I have had to submit patches to conda-forge to unbreak `anaconda-client` before so as its SemVer nature is not checked explicitly, and so securing reproducible environments for it is rather important.
* The update is easy:
   1. Update the `environment.yml` and then run `bash lock.sh`.
   2. Commit the updated `environment.yml` and `conda-lock.yml` to version control and make a new PR.
   3. If the upload tests pass then this new environment can be released to everyone.
* Using a lock file means that no environment solve is done during the build. It was already done when the lock file was created, so to create the environment `micromamba` just needs to download each binary listed in the lock file. (Of course, as this environment is so simple this saves very little time, but it is true in general regardless of the size of the environment.)
* This (hopefully!) means that the SemVer nature of this project's versioning can be leaned into and making new releases is viewed as a trivially verifiable and fast.


lock.sh
```diff
@@ -0,0 +1,16 @@
...
+# Don't need to ensure emulation possible on non-x86_64 platforms at the
+# `docker run` level as the platform is specified in the conda-lock command.
+docker run \
+    --rm \
+    --volume "${PWD}":/work \
+    --workdir /work \
+    "${BUILD_IMAGE}" \
+    /bin/bash -c "\
+        micromamba install --yes --channel conda-forge conda-lock && \
+        conda-lock lock --micromamba --platform linux-64 --file environment.yml"
```
matthewfeickert, line 16: For maintainability I would like to know if this Bash script is clear enough as to what is happening and why (hopefully the name `lock.sh` helps it a bit as well).

Dockerfile
```diff
@@ -8,6 +8,8 @@ SHELL [ "/bin/bash", "-c" ]
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
+# lock file created by running lock.sh in the top level of the repository
+COPY --chown=mambauser conda-lock.yml /conda-lock.yml
```
matthewfeickert, line 12: The lock file is now under version control so we have it accessible to us at the Docker image build time to be able to `COPY` in.

lock.sh
matthewfeickert, line 16: The default user for `mambaorg/micromamba:1.4.9-bullseye-slim` is a non-root user with id `1000` so the generated `conda-lock.yml` is under the ownership of the user who ran the `lock.sh` script. So you don't have to mess around with getting a file owned by `root` and then needing to do permission and ownership changes on it.

bsipocz:
Do you, or anyone else know what this  means on a PR? Is the action configured improperly?

Screenshot 2023-08-14 at 18 37 08


Also, we haven't stated preferred dev workflow, but I always went with the assumption of fork's feature branch+PR is the preferred way rather than keeping feature branches on the main repo. Could we do that in the future?

matthewfeickert:
> Do you, or anyone else know what this means on a PR? Is the action configured improperly?

Things are working as intended. "deployed to remove-old-wheels" means that the `remove-old-wheels` environment (added in PR https://github.com/scientific-python/upload-nightly-action/pull/12) was used in the workflow.

This bit that uses the environment

https://github.com/scientific-python/upload-nightly-action/blob/2b318b62a3436c8e05b7ed8d169a6feff8e9b537/.github/workflows/ci.yml#L56-L62

was added in PR #27 because you asked for the test upload to be removed in the same workflow it was uploaded in (https://github.com/scientific-python/upload-nightly-action/pull/27#discussion_r1281151212).

> I always went with the assumption of fork's feature branch+PR is the preferred way rather than keeping feature branches on the main repo. Could we do that in the future?

No, as the CI needs to run

https://github.com/scientific-python/upload-nightly-action/blob/2b318b62a3436c8e05b7ed8d169a6feff8e9b537/.github/workflows/ci.yml#L17-L21

and the CI requires secrets that are repo/org specific.

https://github.com/scientific-python/upload-nightly-action/blob/2b318b62a3436c8e05b7ed8d169a6feff8e9b537/.github/workflows/ci.yml#L54

https://github.com/scientific-python/upload-nightly-action/blob/2b318b62a3436c8e05b7ed8d169a6feff8e9b537/.github/workflows/ci.yml#L76

There is [`pull_request_target` triggers](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) but that also [introduces security threats](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) and so should be avoided unless there is a very good reason to use otherwise.

I see no problems in having branches other than `main` exist.

matthewfeickert:
This PR would have prevented Issue #43 from affecting users. @Carreau can you review?

Carreau (approved):

MridulS (approved):
This looks good, thanks @matthewfeickert!

We have 2 reviews, not sure if we should wait for more?
If not I can merge this and maybe we can do a 0.2.0 release?

tupui:
Just adding a note here. If we start to use lock files, we need to be careful when integrating changes. There have been cases of supply chain attack using lock files. It's very easy to miss a change in a lock file as these are usually very large and people rarely check the diffs.

bsipocz:
Does it have to be a full on conda lock, cannot we simply stay within pip land and use a much smaller file there?

matthewfeickert:
> Just adding a note here. If we start to use lock files, we need to be careful when integrating changes. There have been cases of supply chain attack using lock files. It's very easy to miss a change in a lock file as these are usually very large and people rarely check the diffs.

@tupui Can you elaborate? If you mean by injection of malicious software by manually editing the lock file then this is checked in two ways:

1. Only allowing incremental updates and not entire re-locks by outside contributors.
2. You can also check the file by checking out the branch and rerunning `lock.sh`. The diff should immeditaltey show if there is anything suspicious.

> Does it have to be a full on conda lock, cannot we simply stay within pip land and use a much smaller file there?

Anaconda does not support `anaconda-client` on PyPI, so no. `pip-compile` is not going to be singificanly smaller though, so that wouldn't help anyway.

MridulS:
> Just adding a note here. If we start to use lock files, we need to be careful when integrating changes. There have been cases of supply chain attack using lock files. It's very easy to miss a change in a lock file as these are usually very large and people rarely check the diffs.

We can also just make github run the lock.sh script and ask it to resolve and generate the lock file, but it would probably be a bit overkill. I don't really think there would be too much activity/changes in lock file here.

matthewfeickert:
> If not I can merge this and maybe we can do a 0.2.0 release?

@MridulS, yes once @tupui and @bsipocz respond (assuming they don't have any further concerns) then I would very much like to get this merged and to get a `v0.2.0` out.

matthewfeickert:
Gentle ping to @tupui and @bsipocz (before it gets to be Friday night in Europe). If there are no further comments by end of the day US time I'm going to merge this as we already have 2 approving reviews from @Carreau and @MridulS. If there are additional comments later we can always iterate.

tupui:
Sorry all good 👍 I just wanted to bring that up.

bsipocz:
It's an ok from me too. But would advise against a Friday night release.

PR #36, opened 2023-08-15
ENH: adding list of packages to exclude from cleanup

### LINKED ISSUE #34: exclude openblas-libs from the cleanup job
The files at https://anaconda.org/scientific-python-nightly-wheels/openblas-libs/files are not meant to be cleaned up automatically. We generate those tarballs (which are not conda packages) for long-term use by SciPy and NumPy. Is there a way to exclude them from https://github.com/scientific-python/upload-nightly-action/blob/main/.github/workflows/remove-wheels.yml ? If not we should go back to keeping them at https://anaconda.org/multibuild-wheels-staging/openblas-libs/

CC @honno, @rgommers

bsipocz:
This is to add a file to host the agreed upon exceptions for the cleanups. Neither the number version nor the date based cleanup will run for these.

CI is expected to fail as 
Fixes #34 

cc @mattip

### DIFF
--- a/.github/workflows/remove-wheels.yml
+++ b/.github/workflows/remove-wheels.yml
@@ -51,9 +51,11 @@ jobs:
 
       - name: Query package index for packages
         run: |
+          curl https://raw.githubusercontent.com/scientific-python/upload-nightly-action/main/packages-ignore-from-cleanup.txt --output packages-ignore-from-cleanup.txt
           anaconda show "${ANACONDA_USER}" &> >(grep "${ANACONDA_USER}/") | \
               awk '{print $1}' | \
-              sed 's|.*/||g' > package-names.txt
+              sed 's|.*/||g' | \
+              grep -vf packages-ignore-from-cleanup.txt > package-names.txt
 
       - name: Remove old uploads to save space
         run: |
--- a/packages-ignore-from-cleanup.txt
+++ b/packages-ignore-from-cleanup.txt
@@ -1,0 +1,1 @@
+openblas-libs


.github/workflows/remove-wheels.yml
```diff
@@ -51,9 +51,11 @@ jobs:
 
       - name: Query package index for packages
         run: |
+          curl https://raw.githubusercontent.com/scientific-python/upload-nightly-action/main/packages-ignore-from-cleanup.txt --output packages-ignore-from-cleanup.txt
```
matthewfeickert, line 54: If the checkout action is used instead then you can test versions on branches prior to merge. So can just throw

https://github.com/scientific-python/upload-nightly-action/blob/2b318b62a3436c8e05b7ed8d169a6feff8e9b537/.github/workflows/ci.yml#L24

at the top of `steps` (or anywhere in `steps` before this).

.github/workflows/remove-wheels.yml
```diff
@@ -51,9 +51,11 @@ jobs:
...
           anaconda show "${ANACONDA_USER}" &> >(grep "${ANACONDA_USER}/") | \
               awk '{print $1}' | \
-              sed 's|.*/||g' > package-names.txt
+              sed 's|.*/||g' | \
+              grep -vf packages-ignore-from-cleanup.txt > package-names.txt
```
matthewfeickert, line 58: Small suggestion: Use the full option names of CLI APIs in CI workflows so that tired people late at night don't have to run `man` or `help` to refresh their memories

```suggestion
              grep --invert-match --file packages-ignore-from-cleanup.txt > package-names.txt
```

matthewfeickert (changes requested):
Two small suggestions to make is easier for future packages to be tested (if needed). Otherwise thanks and looks good.

matthewfeickert:
The reason the CI is failing is described in https://github.com/scientific-python/upload-nightly-action/pull/32#issuecomment-1678400638

.github/workflows/remove-wheels.yml
bsipocz: This was deliberate, given this is the only file needed from the repo, I opted for the curl rather than using an action that potentially need to be version updated, etc.

.github/workflows/remove-wheels.yml
bsipocz: sure. Though `-vf` one is so much a muscle memory, I assume for anyone using shell scripting, that changing it to `--invert-match` I would be the one needing to check the help.

.github/workflows/remove-wheels.yml
matthewfeickert: >  opted for the curl rather than using an action that potentially need to be version updated, etc.

Can you explain more? This doesn't make sense to me. The actions for this workflow are already maintained automatically by Dependabot by `.github/dependabot.yml`. The clone will be trivially fast.

.github/workflows/remove-wheels.yml
matthewfeickert: Using full option names also keeps thing from falling into the alphabet soup of options (here I'm thinking of something like `tar` where using the UNIX-style option flags means that I'm constantly reaching for `man tar` or `cheat tar`).

.github/workflows/remove-wheels.yml
bsipocz: dependabot generating the noise would still be noise 🤷‍♀️ 

And the remove logic is not tested anywhere, so I still don't think it's necessary to do the clone at all, and would otherwise just take this file from main.

.github/workflows/remove-wheels.yml
mattip: I think these command line options are so commonly used that it is fine to keep the short version. If the command used more esoteric grep options the long version would be more helpful, but like @bsipocz I would have to think twice if I saw --invert-match, I didn't even know that was the long version of `-v`.

jarrodmillman (approved):

mattip:
Perhaps the retention policy should be mentioned in the README.md as well?

bsipocz:
Given the one approval, and the discussion on the rest, I go ahead and merge this now.

PR #47, opened 2023-11-29
ENH: Add optional `anaconda_nightly_upload_organization` argument

skoudoro:
This is a follow up and fixes #46

This PR goal is to allow other Scientific Python projects to use this github action.

Please, let me know what you think, and I will be happy to rename the variables if needed.

Thanks  

**Squash and merge commit**

```
* Add optional `anaconda_nightly_upload_organization` and
  `anaconda_nightly_upload_labels` arguments to allow for packages
  to be uploaded to custom registires with the GitHub Action.
* Add docs to README providing an example of how to upload to an
  alternative registry.
```

PR #54, opened 2024-01-24
CI: Add test for uploads with multiple labels

matthewfeickert:
Add test uploads for non-main label 'test' and for multiple labels in a comma seperated list to compliment the feature added in PR #47.

This is an example of things working as intended prior to the cleanup step removing the test package.

pre-clean-up

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
           artifacts_path: dist
           anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
 
-      - name: Build a wheel and a sdist
+      - name: Build v0.0.2 wheel and sdist
         run: |
           # Bump version to avoid wheel name conflicts
           sed -i 's/0.0.1/0.0.2/g' tests/test_package/pyproject.toml
@@ -67,7 +67,7 @@ jobs:
           anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
           anaconda_nightly_upload_labels: test
 
-      - name: Build a wheel and a sdist
+      - name: Build v0.0.3 wheel and sdist
         run: |
           # Bump version to avoid wheel name conflicts
           sed -i 's/0.0.2/0.0.3/g' tests/test_package/pyproject.toml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Build a wheel and a sdist
         run: |
-          PYTHONWARNINGS=error,default::DeprecationWarning python -m build --outdir ./dist tests/test_package
+          python -m build --outdir ./dist tests/test_package
 
       - name: Verify the distribution
         run: twine check --strict dist/*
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -53,6 +53,34 @@ jobs:
           artifacts_path: dist
           anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
 
+      - name: Build a wheel and a sdist
+        run: |
+          # Bump version to avoid wheel name conflicts
+          sed -i 's/0.0.1/0.0.2/g' tests/test_package/pyproject.toml
+          rm ./dist/*
+          python -m build --outdir ./dist tests/test_package
```
matthewfeickert, line 61: As attempting to apply multiple labels through repeated upload of the same package will fail as the package already exists, bump the version number of the test package to generate multiple packages that can be uploaded with different labels.

matthewfeickert (commented):
If we wanted to also check what labels are uploaded as a validation step in `cleanup` before removing them we could add something like

```bash
curl --silent https://api.anaconda.org/package/scientific-python-nightly-wheels/test-package | \
  jq -r '.releases[].version' > package-versions.txt
for package_version in $(cat package-versions.txt); do
  echo "# scientific-python-nightly-wheels/test-package version ${package_version} labels:"
  curl --silent https://api.anaconda.org/package/scientific-python-nightly-wheels/test-package | \
    jq -r '.files[] | \
    select( .version == "${package_version}" )' | \
    jq -r '.labels'
done
```

but I don't think that's strictly necessary, so I've left this out.

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Build a wheel and a sdist
         run: |
-          PYTHONWARNINGS=error,default::DeprecationWarning python -m build --outdir ./dist tests/test_package
+          python -m build --outdir ./dist tests/test_package
```
matthewfeickert, line 39: Removing this as I had added in previously from another workflow I have, but we don't care if there are deprecation warnings in the package build tools, as we just need an artifact to upload.

bsipocz: that assumes that the command will keep working as is, which I think is a valid assumption (or will be a super simple fix many years down the line), so 👍

.github/workflows/ci.yml
```diff
@@ -53,6 +53,34 @@ jobs:
           artifacts_path: dist
           anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
 
+      - name: Build a wheel and a sdist
```
bsipocz, line 56: Too bad it's not possible to merge the `run` and `uses` steps into one, so maybe just change the names so it's not repeated, which may be helpful when quick looking the code
```suggestion
      - name: Build v0.0.2 wheel and sdist
```

.github/workflows/ci.yml
```diff
@@ -53,6 +53,34 @@ jobs:
...
+          python -m build --outdir ./dist tests/test_package
+
+      - name: Test upload with non-main label
+        uses: ./
+        with:
+          artifacts_path: dist
+          anaconda_nightly_upload_token: ${{ secrets.UPLOAD_TOKEN }}
+          anaconda_nightly_upload_labels: test
+
+      - name: Build a wheel and a sdist
```
bsipocz, line 70:
```suggestion
      - name: Build v0.0.3 wheel and sdist
```

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Build a wheel and a sdist
```
bsipocz, line 37:
```suggestion
      - name: Build v0.0.1 wheel and sdist
```

bsipocz (approved):
I only have extreme nitpicks, and also wondered about the same thing you wrote whether the labels need to be verified. I don't necessarily think, so would say go ahead and merge this without that check.

.github/workflows/ci.yml
matthewfeickert: Yeah, the main reason I removed this is that I've had this check in other work flows and this has ended up causing things to fail for reasons unrelated to `build` itself (e.g. a build backend deprecating something for the future but that warning still getting caught). I agree that if `build` ends up having issues though we will not be the first to catch it. :)

PR #59, opened 2024-01-25
CI: Group dependabot updates

matthewfeickert:
* Group dependabot updates to reduce the number of PRs.
   - c.f. sp-repo-review GH212: Require GHA update grouping https://learn.scientific-python.org/development/guides/gha-basic/#GH212

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,5 +5,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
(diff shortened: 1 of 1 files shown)


matthewfeickert:
As this is simply reducing the number of PRs that will be made by Dependabot I am going to assume that is fine with everyone, and so instead of asking for a review I'm going to just merge.

Apologies in advance if people would prefer to always have 1 other review. I'm happy to do so in the future.

PR #64, opened 2024-02-15
fix: Guard against removing package while uploading to it

### LINKED ISSUE #57: Upload of package version of same name causes error and package removal
Anaconda Cloud seems to have changed how they handle uploads of packages where there is already a package of the same name and version on Anaconda Cloud. In the past, this simply just resulted in the artifact being uploaded overwriting the previous artifact. `networkx` is one of the core packages that used this approach. However, I noticed that [`networkx`'s nightly workflow](https://github.com/networkx/networkx/blob/d6569adf7c224a9c85c0e4e4c92435f772f73582/.github/workflows/nightly.yml) is failing to do so

image

```
...
File type is "Standard Python"
Extracting standard python attributes for upload
Creating package "networkx"
Creating release "3.3rc0.dev0"
Uploading file "scientific-python-nightly-wheels/networkx/3.3rc0.dev0/networkx-3.3rc0.dev0-py3-none-any.whl"
Warning:  Distribution "networkx-3.3rc0.dev0-py3-none-any.whl" already exists. Removing.
Error:  ("release version='3.3rc0.dev0' does not exist", 404)
##[debug]Docker Action run completed with exit code 1
##[debug]Finishing: Upload nighlty wheel
```

(The success after failure is due to the package being removed and so there is no conflict the next upload.)

This started happening on 2024-01-04 for `networkx`, and was not happening before that. For example, this 2024-01-02 upload ran successfully

```
+ anaconda --token *** upload --force --user scientific-python-nightly-wheels dist/networkx-3.3rc0.dev0-py3-none-any.whl
...
Using "scientific-python-nightly-wheels" as upload username
Processing "dist/networkx-3.3rc0.dev0-py3-none-any.whl"
Detecting file type...
File type is "Standard Python"
Extracting standard python attributes for upload
Creating package "networkx"
Creating release "3.3rc0.dev0"
Uploading file "scientific-python-nightly-wheels/networkx/3.3rc0.dev0/networkx-3.3rc0.dev0-py3-none-any.whl"
Warning:  Distribution "networkx-3.3rc0.dev0-py3-none-any.whl" already exists. Removing.

  0%|          | 0.00/1.59M [00:00<?, ?B/s]
1.59MB [00:00, 5.26MB/s]                   
Upload complete

standard python located at:
  https://anaconda.org/scientific-python-nightly-wheels/networkx
```

We know that as we're using a [lock file](https://github.com/scientific-python/upload-nightly-action/blob/6f1e2b99c3148480df25d296c27ae58c446ad5d0/conda-lock.yml)

https://github.com/scientific-python/upload-nightly-action/blob/6f1e2b99c3148480df25d296c27ae58c446ad5d0/cmd.sh#L57-L60

that nothing has changed from the `upload-nightly-action` action side, and so this has to be on the Anaconda Cloud side.

cc as a heads up to @MridulS @jarrodmillman

matthewfeickert:
Resolves #57 

* On Anaconda Cloud as of 2024-01-04, if a wheel is being uploaded to a package, but the package only has one wheel in it and is of the same name as the uploaded wheel, Anaconda Cloud will overwrite the file by _removing_ the file from the package index. However, when this happens it removes the entire package, and then the wheel that is in the process of being uploaded has no destination and the upload fails. To guard against this, ensure for each package that has a wheel being uploaded if:
   - there is only one release for the package
   - and only 1 file for that release
   - and the upload target wheel has the same name as the file
   - that the file (and so the package) is removed in advance of the
     upload.

* To make filtering names and versions from wheels easier, add a `get_wheel_name_version` function that uses as regex to lazily capture the package name as well as the version and then return these.
   - Examples of this working:
     * `"matplotlib-3.9.0.dev0-pp39-pypy39_pp73-win_amd64.whl"`
       `matplotlib` `3.9.0.dev0`

     * `"scikit_learn-1.5.dev0-cp39-cp39-win_amd64.whl"`
       `scikit_learn` `1.5.dev0`

     * `"scipy-openblas64-0.3.26.186-py3-none-macosx_10_9_x86_64.whl"`
       `scipy-openblas64` `0.3.26.186`

     * `"awkward_cpp-29-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"`
       `awkward_cpp` `29`

     * `"awkward-2.6.1-py3-none-any.whl"`
       `awkward` `2.6.1`

* As this requires `curl` and `jq` also add these to the environment and relock.

* Add a extra upload in CI to thest this behavior.

In the future there will hopefully be a fix on Anaconda Cloud's side given https://github.com/Anaconda-Platform/anaconda-client/issues/702 but we need a solution now.

### DIFF
--- a/cmd.sh
+++ b/cmd.sh
@@ -67,6 +67,70 @@ micromamba activate upload-nightly-action
 # trim trailing slashes from $INPUT_ARTIFACTS_PATH
 INPUT_ARTIFACTS_PATH="${INPUT_ARTIFACTS_PATH%/}"
 
+get_wheel_name_version() {
+    local wheel_name="$1"
+    if [[ "${wheel_name}" =~ ^([[:alnum:]_-]+)-([0-9][^-]+)-(.+)$ ]]; then
+        # return the package_name and version_number
+        local return_values=("${BASH_REMATCH[1]}" "${BASH_REMATCH[2]}")
+        echo "${return_values[@]}"
+    else
+        echo "The wheel name ${1} does not follow the PEP 491 spec (https://peps.python.org/pep-0491/) and is invalid."
+        return 1
+    fi
+}
+
+# get the unique package names from all the wheels
+package_names=()
+for wheel_path in "${INPUT_ARTIFACTS_PATH}"/*.whl; do
+  # remove the INPUT_ARTIFACTS_PATH/ prefix (including the /)
+  wheel_name="${wheel_path#${INPUT_ARTIFACTS_PATH}/}"
+  read -r package_basename_prefix _ <<< "$(get_wheel_name_version ${wheel_name})"
+  package_names+=("${package_basename_prefix}")
+done
+package_names=($(tr ' ' '\n' <<< "${package_names[@]}" | sort --unique | tr '\n' ' '))
+
+# If the package version doesn't exist on the package index then there will
+# be no files to overwrite and the package can be uploaded safely.
+# If the package version exists, is the only version on the package index,
+# and only has one distribution file, then that package needs to be removed
+# from the index before a wheel of the same name can be uploaded again.
+# c.f. https://github.com/Anaconda-Platform/anaconda-client/issues/702
+
+for package_basename_prefix in "${package_names[@]}"; do
+  # normalize package_name to use '-' as delimiter
+  package_name="${package_basename_prefix//_/-}"
+
+  number_releases=$(curl --silent https://api.anaconda.org/package/"${ANACONDA_ORG}/${package_name}" | \
+    jq -r '.releases' | \
+    jq length)
+
+  if [ "${number_releases}" -eq 1 ]; then
+    # get any wheel for the package (they should all have the same version)
+    wheel_path=$(find "${INPUT_ARTIFACTS_PATH}" -name "${package_basename_prefix}-*.whl" -print -quit)
+    wheel_name="${wheel_path#${INPUT_ARTIFACTS_PATH}/}"
+    read -r _ package_version <<< "$(get_wheel_name_version ${wheel_name})"
+
+    number_files=$(curl --silent https://api.anaconda.org/release/"${ANACONDA_ORG}/${package_name}/${package_version}" | \
+      jq -r '.distributions' | \
+      jq length)
+
+    if [ "${number_files}" -eq 1 ]; then
+      distribution_name=$(curl --silent https://api.anaconda.org/release/"${ANACONDA_ORG}/${package_name}/${package_version}" | \
+        jq -r '.distributions[].basename')
+
+      if [ "${wheel_name}" = "${distribution_name}" ]; then
+        echo -e "\n# ${distribution_name} is the only distribution file uploaded for the package https://anaconda.org/${ANACONDA_ORG}/${package_name}"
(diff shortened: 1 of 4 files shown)


cmd.sh
```diff
@@ -67,6 +67,58 @@ micromamba activate upload-nightly-action
 # trim trailing slashes from $INPUT_ARTIFACTS_PATH
 INPUT_ARTIFACTS_PATH="${INPUT_ARTIFACTS_PATH%/}"
 
+# TODO: First get a list of just the package names so
+# this doesn't have to get run for every wheel
```
matthewfeickert, line 71: Need to do this first before this can be ready for review so that this loop isn't run for _every_ wheel.

cmd.sh
```diff
@@ -67,6 +67,64 @@ micromamba activate upload-nightly-action
...
+# If the package version doesn't exist on the package index then there will
+# be no files to overwrite and the package can be uploaded safely.
+# If the package version exists, is the only version on the package index,
+# and only has one distribution file, then that package needs to be removed
+# from the index before a wheel of the same name can be uploaded again.
+# c.f. https://github.com/Anaconda-Platform/anaconda-client/issues/702
+
+for package_basename_prefix in "${package_names[@]}"; do
+  # normalize package_name to use '-' as delimiter
+  package_name="${package_basename_prefix//_/-}"
```
matthewfeickert, line 90: Need to check the definition of what a normalized name is for wheels.

matthewfeickert: c.f. https://github.com/pypa/setuptools/issues/4214 for reference.

matthewfeickert:
@kenodegard @dholth as you know the Anaconda Cloud web APIs better than I do, if you would have time to review the changes to `cmd.sh` and let me know if there are better API calls to be making that would be excellent and much appreciated.

matthewfeickert:
@scientific-python/nightly-wheels-developers If it would be possible to get a review early this week it would be great to get this in so that I can create a new patch release with this.

cmd.sh
```diff
@@ -67,6 +67,70 @@ micromamba activate upload-nightly-action
...
+
+  number_releases=$(curl --silent https://api.anaconda.org/package/"${ANACONDA_ORG}/${package_name}" | \
+    jq -r '.releases' | \
+    jq length)
+
+  if [ "${number_releases}" -eq 1 ]; then
+    # get any wheel for the package (they should all have the same version)
+    wheel_path=$(find "${INPUT_ARTIFACTS_PATH}" -name "${package_basename_prefix}-*.whl" -print -quit)
+    wheel_name="${wheel_path#${INPUT_ARTIFACTS_PATH}/}"
+    read -r package_basename_prefix package_version <<< "$(get_wheel_name_version ${wheel_name})"
```
matthewfeickert, line 111:
```suggestion
    read -r _ package_version <<< "$(get_wheel_name_version ${wheel_name})"
```

`package_basename_prefix` isn't used later in the script, so can just replace it with a filler here.

matthewfeickert: Fixed in rebase.

matthewfeickert:
@scientific-python/nightly-wheels-developers Ping again on this, as given that apparently merging now requires a minimum of 1 review this can't get merged until someone reviews.

jarrodmillman (approved):

matthewfeickert:
Thanks @jarrodmillman! I'm going to merge and then prep a new minor release as this add behavior without introducing breaking changes.

PR #79, opened 2024-04-17
CI: fix spelling of test package name

bsipocz:
`python -m build --outdir ./dist tests/test_package` now creates `test_package-0.0.1.tar.gz` and `test_package-0.0.1-py3-none-any.whl` thus this PR fixes the spelling for the CI to pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
         run: twine check --strict dist/*
 
       - name: List contents of sdist
-        run: python -m tarfile --list dist/test-package-*.tar.gz
+        run: python -m tarfile --list dist/test_package-*.tar.gz
 
       - name: List contents of wheel
         run: python -m zipfile --list dist/test_package-*.whl


matthewfeickert (approved):
Thanks @bsipocz. This is (I'm assuming) a change in how `hatchling` does things, as `build` shouldn't control this, so for posterity and my own education I'll check on this (unless @henryiii knows of the top of his head).

matthewfeickert:
Ah yes. In [`hatchling` `v1.22.0`](https://github.com/pypa/hatch/releases/tag/hatchling-v1.22.0)

> Metadata for the `wheel` target now defaults to the `PKG-INFO` metadata within source distributions

Nice.

henryiii:
I thought hatchling has normalized SDist names for among time? Setuptools *just* changed to do it too, but I though hatching basically always did it?

henryiii:
Yes, it’s setuptools: https://github.com/scientific-python/upload-nightly-action/blob/a5d8097406eb99827061cc20d199db55e82eef6b/tests/test_package/pyproject.toml#L2

All the other build backends have done this since nearly the beginning.

matthewfeickert:
Oh right. I just totally forgot that I had to use `setuptools` in PR #27 because https://github.com/anaconda/anaconda-client/issues/676 blocks this. I should see if this is still the case.

PR #81, opened 2024-06-03
Using hashes for all actions

bsipocz:


.github/workflows/ci.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     if: github.repository == 'scientific-python/upload-nightly-action'
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29  # v4.1.6
```
Carreau, line 24: @tupui you said that maybe we had to use hash:hash ?

matthewfeickert (approved):

.github/workflows/ci.yml
matthewfeickert: I'm slightly less concerned about these for projects that are known. But maybe I don't understand @tupui's point.

tupui: I am trying to find where I read that 😅 Still digging but yes the hack was that if you had a branch named as the hash, then it would be picked up. In the article they were mentioning some protections from GitHub but there was still a way to do the hack IIRC.

bsipocz:
Given that we didn't find the syntax of specifying that the hash is a hash, let's go ahead with this as is. I tried to create 40 long hexa branches but couldn't do that on the GH web interface, nor I could push such a branch from a local checkout. 

Screenshot 2024-06-05 at 17 44 11

stefanv (approved):

stefanv:
Normal for CI to fail?

bsipocz:
> Normal for CI to fail?

I believe it is expected to fails for all but @matthewfeickert

matthewfeickert:
> > Normal for CI to fail?
> 
> I believe it is expected to fails for all but @matthewfeickert

No, it has nothing to do with my account, it just will always fail if a PR from a fork as the CI requires secrets that are repo/org specific, as described in https://github.com/scientific-python/upload-nightly-action/pull/32#issuecomment-1678400638

matthewfeickert:
> Given that we didn't find the syntax of specifying that the hash is a hash, let's go ahead with this as is. I tried to create 40 long hexa branches but couldn't do that on the GH web interface, nor I could push such a branch from a local checkout.

SGTM, so let's merge. :shipit:

bsipocz:
> No, it has nothing to do with my account, it just will always fail if a PR from a fork 

Oh, ok. I haven't realized that you didn't open your PRs from your fork, just noticed that those all have the green tickmarks while the rest are crossed out.

PR #88, opened 2024-08-28
docs: Correct token url to user account

matthewfeickert:
Address typo noticed in https://github.com/apache/arrow/issues/40216#issuecomment-2315175755.

Users of the action should generate an access token from their own Anaconda Cloud account, not from the Scientific Python organisation's account.


MridulS (approved):
Thanks!

PR #91, opened 2024-09-19
ENH: Use pixi to run action as composite

### LINKED ISSUE #71: Document that the action requires to run on ubuntu
I naively added the action to our cibuildwheel matrix (for the shapely wheels), expecting to upload the wheels from each OS-specific build directly. 
However, this does not work because the action apparently requires a linux image: *"Container action is only supported on Linux"* (see eg https://github.com/shapely/shapely/actions/runs/8047467718/job/21976750565). So you need to gather all wheel for the different architectures and upload them from another job running on ubuntu.

matthewfeickert:
Resolves #71 

## Description

* Use `pixi` to manage the environment fully, which removes the needs for `micromamba` and a Linux container, allowing the action to be run as a 'composite' action.
   - This allows the action to be used on macOS runners in addition to Linux.
   - Add `pixi` project configuration (`pixi.toml`) and `pixi` lock file (`pixi.lock`).
   - Add 'osx-64' and 'osx-arm64' platforms to pixi environment solves to allow for use on Linux and macOS runners.
   - Use `prefix-dev/setup-pixi` GitHub action to setup `pixi` for the action.
   - Explicitly enable the 'locked' option for `prefix-dev/setup-pixi`.
   - Don't allow for post job cleanup to avoid errors if the action is run multiple times in a workflow (this is a downside of no longer being in a Linux container).
   - Make `cmd.sh` executable (`chmod +x`).
* Remove use of micromamba in the upload script.
* Remove the conda-lock files and the Docker based locking workflow.
* Remove `Dockerfile`.
* Note in the README that the runner used can be either Linux or macOS.
* Update version numbers of action to `v0.5.0`.

* Rename script from `cmd.sh` to `upload_wheels.sh`.
   - As the script is no longer the CMD of a Linux container but is now a standalone Bash script, having a more descriptive name is preferable.

> [!NOTE] 
> GitHub Actions takes care of cleaning up the environment after each run, so you are not responsible for manually unsetting all environment variables between steps to avoid secret leakage.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mambaorg/micromamba:1.4.9-bullseye-slim
+FROM mambaorg/micromamba:1.5.9-bookworm-slim
```
stefanv, line 1: Is micromamba still a thing? Perhaps this should now be https://quay.io/repository/condaforge/miniforge3

default_linux-64_conda_spec.txt
```diff
@@ -0,0 +1,68 @@
+# Generated by `pixi project export`
```
stefanv, line 1: Needed when pixi.lock exists?

pixi.toml
```diff
@@ -0,0 +1,14 @@
+[project]
+authors = ["Matthew Feickert <matthew.feickert@cern.ch>"]
+channels = ["conda-forge"]
```
stefanv, line 3: :+1:

Dockerfile
```diff
@@ -8,8 +8,9 @@ SHELL [ "/bin/bash", "-c" ]
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-# lock file created by running lock.sh in the top level of the repository
-COPY --chown=mambauser conda-lock.yml /conda-lock.yml
+# explicit specification file file created from pixi.lock by running
+# 'pixi run export' in the top level of the repository
+COPY --chown=mambauser default_linux-64_conda_spec.txt /default_linux-64_conda_spec.txt
```
stefanv, line 13: Maybe modify the cmd to create environment directly using pixi?

stefanv (commented):
:sunglasses:

bsipocz:
Nitpicking: @matthewfeickert could we follow a fork-based workflow here in the future, so the main repo won't be peppered with feature branches?

Dockerfile
matthewfeickert: I think so. I may be wrong here, but I believe that `mamba`/`micromamba` are still developed together on https://github.com/mamba-org/mamba.

Dockerfile
matthewfeickert: That would definitely be easier, but I assumed that people would be more receptive to an incremental change of having the runtime environment stay with `micromamba` at first. Switching everything to `pixi` solves the error of course too.

pixi.toml
```diff
@@ -0,0 +1,14 @@
+[project]
+authors = ["Matthew Feickert <matthew.feickert@cern.ch>"]
```
matthewfeickert, line 2: Is there a Scientific Python org email that we can use here instead? `pixi` generates this information from your Git credentials automatically during `pixi init`.

bsipocz: Are there alias email addresses for e.g. a discorse channel? That case it would be a fully transparent address. 

Also, the docs says that authors is optional, so maybe leaving it empty or as `Scientific Python`, as in the Licence, without an email address could work, too.

bsipocz:
> I also don't find a transient feature branch terrible

The issue is that some of the branches are not exactly transient, but are around for a while when part of some experimenting or draft PR.

pixi.toml
matthewfeickert, line 2:
```suggestion
authors = ["Scientific Python"]
```

bsipocz:
Re CI: I think it's OK to separate the current workflow into two, one that runs all the anaconda upload/cleanup checks, and one that does everything else. Then we can run most of the tests from forks, too, and in practice those would cover most of the PR cases anyway.

matthewfeickert:
> Re CI: I think it's OK to separate the current workflow into two, one that runs all the anaconda upload/cleanup checks, and one that does everything else. Then we can run most of the tests from forks, too, and in practice those would cover most of the PR cases anyway.

@bsipocz I don't think that will work for the reasons I describe in the two previous responses I've linked, as you'll need secrets to run the upload and a secret to delete the files and neither can be run on a fork safely. But can you make an Issue to describe more what you have in mind here?

pixi.toml
stefanv: We can make a discourse group, and attach a x@discuss.scientific-python.org address to it, or I can make an alias for x@scientific-python.org and redirect it wherever. We don't have the concept of forward-email-into-discord atm.

Dockerfile
stefanv: I'd say go for it, if you like the idea.

tupui:
Cool 👍 Just a fly by comment. What do you think about uv otherwise? Looks like this is going to be the thing sooner than later.

bsipocz:
> I don't think that will work for the reasons I describe in the two previous responses I've linked, as you'll need secrets to run the upload and a secret to delete the files and neither can be run on a fork safely.

Yes, my point is to separate those that need the secret in a separate workflow/job, so 1) tests can pass on forks with the acknowledgement that the secret-requiring ones are skipped.

But also, something is broken with the current setup as e.g. #86 failed even though it run from a feature branch here.

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
     default: main
 
 runs:
-  using: 'docker'
-  image: 'Dockerfile'
+  using: "composite"
+  steps:
+    - name: Set up pixi
+      uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659  # v0.8.1
```
matthewfeickert, line 30: I need to look at what the [`prefix-dev/setup-pixi`](https://github.com/prefix-dev/setup-pixi) action is actually doing to see if pinning at the hash level actually provides any safeguards. I'll also ask the `pixi` team to review this PR once it is ready to move from draft to review.

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
...
+      with:
+        cache: true
+        cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```
matthewfeickert, line 33: I'm not fully clear on how `github.ref_name` works when the action is being used by other workflows, so this might all be wrong. (Here I'm just copying from the https://github.com/prefix-dev/setup-pixi README).

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
...
+        # Avoid post cleanup errors if action run multiple times
+        post-cleanup: false
```
matthewfeickert, line 35: If the action is run multiple times in the same workflow (like we do in the CI to test) then at the end of the workflow each of the job steps gets cleaned up sequentially. [As](https://github.com/prefix-dev/setup-pixi/blob/6efead024a876e5a222673996735f4972a443a66/action.yml#L58-L61) 

```yaml
  post-cleanup:
    description: |
      If the action should clean up after itself. Defaults to `true`.
      If `true`, the pixi environment, the pixi binary and the rattler files in ~/.rattler and ~/.cache/rattler are removed.
```

this will try to delete the same directory path multiple times and will error when those paths don't exist after the first clean up.

I'd like to find a smarter way around this, as I'm not sure if leaving things around is ideal, but I also don't think it is _bad_.

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
...
+
+    - name: Upload wheels
+      shell: bash
+      env:
+        INPUT_ARTIFACTS_PATH: ${{ inputs.artifacts_path }}
+        INPUT_ANACONDA_NIGHTLY_UPLOAD_ORGANIZATION: ${{ inputs.anaconda_nightly_upload_organization }}
+        INPUT_ANACONDA_NIGHTLY_UPLOAD_TOKEN: ${{ inputs.anaconda_nightly_upload_token }}
+        INPUT_ANACONDA_NIGHTLY_UPLOAD_LABELS: ${{ inputs.anaconda_nightly_upload_labels }}
+      run: |
+        pixi run cmd.sh
```
matthewfeickert, line 45: `cmd.sh` is executable now (post `chmod +x cmd.sh`). We can also rename it to something more descriptive.

cmd.sh
```diff
@@ -78,3 +66,13 @@ anaconda --token "${ANACONDA_TOKEN}" upload \
...
+
+# env cleanup
+unset INPUT_ARTIFACTS_PATH
+unset INPUT_ANACONDA_NIGHTLY_UPLOAD_ORGANIZATION
+unset INPUT_ANACONDA_NIGHTLY_UPLOAD_TOKEN
+unset INPUT_ANACONDA_NIGHTLY_UPLOAD_LABELS
+
+unset ANACONDA_ORG
+unset ANACONDA_TOKEN
+unset ANACONDA_LABELS
```
matthewfeickert, line 78: This is my paranoia of leaking secrets accidentally, but I don't think it is actually needed.

pixi.toml
```diff
@@ -0,0 +1,13 @@
+[project]
+authors = ["Scientific Python"]
+channels = ["conda-forge"]
+description = "Environment for the scientific-python/upload-nightly-action GitHub action"
+name = "upload-nightly-action"
+platforms = ["linux-64", "osx-64", "osx-arm64"]
```
matthewfeickert, line 6: Adding the macOS platforms as though I assume that people will be using this following a `cibuildwheel` workflow on Linux, providing the extra platforms in the lock file allows for this action to be run on macOS runners as well.

matthewfeickert (commented):
> [!NOTE] 
> Though this is passing CI, I'm going to leave this in draft for a bit, to think if there's cleaner and safer ways to do everything, before I move it to 'ready for review' and ask people to spend time on it. That realistically won't be until next week though given that I have work (travel, talks, reviews) and personal deadline-driven tasks looming.
> 
> (I'll also want to rebase things and try to write more clear commit messages for reviewers.)

matthewfeickert:
> What do you think about uv otherwise? Looks like this is going to be the thing sooner than later.

@tupui I like `uv` a lot and I use it on many projects. `uv pip compile` is excellent and I use it in container runtimes where I [don't have control over the Python runtime](https://github.com/usatlas/analysisbase-dask/blob/72c930aca24884fceacf780655c37c62592bfc21/Makefile) given the base image environment (this is actually [one area/thing that `pixi` doesn't really allow you to do](https://github.com/prefix-dev/pixi/discussions/1064) as it needs to control the entire environment &mdash; some caveats there if you export the shell activation though and manually use it.)

That being said, I like `pixi` more in general, and I think for something like this where you want to have hash level reproducibility of the entire stack and not just of the Python packages then `pixi` is a clear winner (well really more just "locked conda installs in an isolated/containerized environment" is the winner).

README.md
```diff
@@ -4,7 +4,8 @@ This is a GitHub Action that uploads nightly builds to the [scientific-python ni
 as recommended in [SPEC4 — Using and Creating Nightly Wheels][].
 
 In a GitHub Actions workflow (`.github/workflows/*.yaml`), use the
-following snippet to upload built wheels to the channel:
+following snippet on a Linux or macOS runner to upload built wheels to the
```
matthewfeickert, line 7: Adding this as we've gotten questions in the past about what runner OS are supported and this PR extends support beyond Linux.

action.yml
matthewfeickert: This still holds true, but worst case scenario we cache for some and not others and I think that's okay.

matthewfeickert (commented):
These are just some final comments in addition to the comments I left for others in my previous self-reviews.

matthewfeickert:
@scientific-python/nightly-wheels-developers this is now ready for review. As I had this in draft before to figure things out I've intentionally squashed and rebased this so that it will be easier to follow my actual intended changes from both the GitHub UI and from a local checkout.

@pavelzw @ruben-arts as you've contributed the most to the [`setup-pixi` action](https://github.com/prefix-dev/setup-pixi) if you're able to review this as well that would be very much appreciated.

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
     default: main
 
 runs:
-  using: 'docker'
-  image: 'Dockerfile'
+  using: "composite"
+  steps:
+    - name: Set up pixi
+      uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659  # v0.8.1
+      with:
```
MridulS, line 31: Should we pin the pixi version? I think it downloads the latest pixi if there is no version specified

action.yml
pavelzw: the setup-pixi action is designed to be backwards compatible (like pixi as well) but there are no guarantees yet.

with a proper dependabot configuration and automatic updates, i would just keep it pinned to the SHA level, though for additional reproducibility

action.yml
pavelzw: it downloads the latest, yes. as there is no proper autoupdate mechanism i personally would keep it at `latest` or suggest you write your own. otherwise you will end up with stale pixi versions in ci

action.yml
pavelzw: currently this is the only workaround. there are some nieche use cases where it actually makes sense to run setup-pixi twice. we might want to open an issue for this

action.yml
```diff
@@ -24,5 +24,22 @@ inputs:
...
+        cache: true
```
matthewfeickert, line 32:
```suggestion
        locked: true
        cache: true
```

Explicitly enforce the default (of `locked` being used) in the event that somehow changes in the future.

c.f. https://github.com/scientific-python/upload-nightly-action/pull/91/files#r1777356588

action.yml
MridulS: > I personally might soften my view here and say that having the latest pixi get installed is probably a good idea / will have no effect given that the lock file is the thing of importance.

I am fine with that too :)

MridulS (approved):

matthewfeickert:
Thanks for reviewing for @scientific-python/nightly-wheels-developers, @MridulS! :pray: If we're able to get one more review from another @scientific-python/nightly-wheels-developers person that would be great, but if no one else reviews by Tuesday (2024-10-01) I'll go ahead and squash and merge.

bsipocz (approved):
Looks good to me, thank you @matthewfeickert for the refactor!

matthewfeickert:
Thanks @bsipocz!

I'm going to do a relock and it there's any changes to a rebase and push and then squash and merge this. I'll then update the `v0.5.0` info in the repo to `v0.6.0` and make a new release on GitHub.

Issue #75, opened 2024-03-08
Request access for PyWavelets


rgommers:
Hi all, it's be great to get access to https://anaconda.org/scientific-python-nightly-wheels/ for [PyWavelets](https://github.com/PyWavelets/pywt/) nightlies.

PyWavelets is a pretty-slow moving project with small wheels, so the extra load should be light. Two reasons why I'm asking for access:
1. PyWavelets is an optional dependency of scikit-image, which is already hosted here - and @jarrodmillman mentioned that it'd be good to have space for the dependency trees of the already-hosted projects (will immediately come in handy with the NumPy 2.0 rebuilds I suspect).
2. We're working on interactive docs and like to use PyWavelets as one of the first packages to have them, since it's small and easy to work on. For that we need Pyodide (`wasm32`) wheels, which can be uploaded to the anaconda.org buckets I believe, but aren't allowed on PyPI/TestPyPI yet.

matthewfeickert:
@rgommers please complete the missing items from the checklist to get started.

* [x] Create an account on [Anaconda Cloud](https://anaconda.org/)
* [x] Reply here with the link to your account (https://anaconda.org/rgommers)
* [ ] Make a new Anaconda Cloud API token with permissions described in https://scientific-python.org/specs/spec-0004/#process-for-adding-new-projects

![image](https://github.com/scientific-python/upload-nightly-action/assets/5142394/697ce89f-4a51-43d0-a26f-2493f4bdce34)

* [ ] Save the API token as a repo GitHub Actions secret named `ANACONDA_ORG_UPLOAD_TOKEN`
* [ ] Direct me (@matthewfeickert) to a `pywavelets` nightly wheel for me to upload to create the package in the registry
* [x] Verify that I have added you to the [`pywavelets` team "Group"](https://anaconda.org/scientific-python-nightly-wheels/groups/pywavelets) on https://anaconda.org/scientific-python-nightly-wheels
* [ ] Verify that I have added the package `scientific-python-nightly-wheels/pywavelets` to the [`pywavelets` team "Group"](https://anaconda.org/scientific-python-nightly-wheels/groups/pywavelets)

[...]

PR #95, opened 2024-09-27
Add a separate action for removing old wheels

### LINKED ISSUE #62: Make Remove Old Wheels workflow seperate GitHub Action
As we're getting requests from different libraries to be added (Issues #29, #45, #50) and recommending to those groups to instead make their own Anaconda Cloud org and use the [`0.3.0` release of the action](https://github.com/scientific-python/upload-nightly-action/releases/tag/0.3.0) to upload to their own org (https://github.com/scientific-python/upload-nightly-action/pull/47#issuecomment-1908716794, https://github.com/scikit-hep/awkward/issues/3011) it would be nice to also not have these teams have to reimpliment their own space saving workflow. 

It would be easy to turn the ["Remove old wheels" workflow](https://github.com/scientific-python/upload-nightly-action/blob/66bc1b6beedff9619cdff8f3361a06802c8f5874/.github/workflows/remove-wheels.yml) into a YAML only (no Docker needed) GitHub Action to pair with the `upload-nightly-action`.

Thoughts?

agriyakhetarpal:
## Description

This PR closes #62; by converting the `.github/workflows/remove-wheels.yml` workflow to a composite action in `remove-wheels/action.yml` that users can trigger in their workflows separately to delete binaries uploaded to the https://anaconda.org/scientific-python-nightly-wheels/ index or other Anaconda.org indices.

This way, users should be able to add, for example:

```yaml
- name: Remove old wheels
  uses: scientific-python/upload-nightly-action/remove-wheels@vx.y.z # or, better, via the commit hash
  with:
    anaconda_nightly_token: ${{ secrets.ANACONDA_NIGHTLY_UPLOAD_TOKEN }} # same as that for uploads (convenience?)
    n_latest_uploads_to_keep: '5'
    anaconda_nightly_upload_organization: 'my-custom-org'
    package_name: 'my-package'
```

or similar variations, to delete all but the last five wheel uploads, which means that this repository does not need to do so and can delegate this responsibility to the workflow files for the packages listed under the https://anaconda.org/scientific-python-nightly-wheels/repo page.

### Changes made

- A new action in `./remove-wheels/action.yml` that accepts an organisation name, a package name, and the number of latest uploads to keep
- The wheel removal logic extracted into a nicer Python script at `remove-wheels/remove_wheels.py` (based on discussions I had with @Cadair)
- Documentation updates in the README

### Checklist

- [x] Resolve the to-do items (some of them have been marked as inline comments in the code for now)
- [x] Add relevant documentation, and adjust headings in the README
- ~~If `.github/workflows/remove-wheels.yml` stays, figure out how to not unintentionally delete wheels without the existence of `packages-ignore-from-cleanup.txt`~~


agriyakhetarpal:
Okay, I've just read some discussion(s) on keeping the removal of old wheels "centralised" in https://github.com/scientific-python/upload-nightly-action/issues/4#issuecomment-1564825980 and in the comments after it. Those made sense to me, so, this PR could be considered in conjunction with that policy – to let SPNW clean up old wheels itself, and let these changes be a "refactor" of the wheel removal logic in a sense, i.e., from being embedded in a workflow to a separate action that gets used in it. This would also ease the removal process around those who want to take up the responsibility of removing their wheels that have been uploaded outside SPNW-managed Anaconda.org indices. It is to be noted that that discussion predates the ones linked in #62, so it is likely the case that requests for easier removals have arisen in more contexts by now.

bsipocz:
Yes, having the cleanup parts moved to a new action and then using it sounds to be the right way to go, that way we expose the functionality for easy reuse while we also keep using it.

matthewfeickert:
I am in transit and won't have time to review or give feedback on this for a few days. My original intent in Issue https://github.com/scientific-python/upload-nightly-action/issues/62 was to make a new repository and have that action be totally independent of this repository. Though, we can certainly get this in here and then just move it elsewhere later if people agree that's a good idea.

agriyakhetarpal:
That sounds great! Thanks for triggering the workflow for me – I shall revisit this either later in the day or early next week. That said, I don't think one can run the workflow from forks, though... I have an account on Anaconda.org already since I'm helping manage the PyWavelets wheels (#75), should I add a dummy package for testing?

agriyakhetarpal:
> to make a new repository and have that action be totally independent of this repository. Though, we can certainly get this in here and then just move it elsewhere later if people agree that's a good idea.

Personally, I would say that related actions like these ones could be grouped together just fine, since one repository is easier to manage, rather than two – easier permissions to deal with, unified documentation in the README, unified dependency management (using `pixi`), etc., so this could fit in the same repository very well. For another example, https://github.com/github/codeql-action contains six actions, with each managed in its own folder.

agriyakhetarpal:
> to make a new repository and have that action be totally independent of this repository. Though, we can certainly get this in here and then just move it elsewhere later if people agree that's a good idea.

Personally, I would say that related actions like these ones could be grouped together just fine, since one repository is easier to manage, rather than two – easier permissions to deal with, unified documentation in the README, unified dependency management (using `pixi`), etc., so this could fit in the same repository very well. For example, https://github.com/github/codeql-action contains six actions, with each managed in its own folder.

README.md
```diff
@@ -18,11 +19,45 @@ jobs:
...
+
+```yml
+jobs:
+  steps:
+    ...
+    - name: Remove old wheels
+      uses: scientific-python/upload-nightly-action/remove-wheels@cantknowhashyet # 0.6.0
+      with:
+        n_latest_uploads: ${{ env.N_LATEST_UPLOADS }}
+        anaconda_nightly_upload_token: ${{secrets.UPLOAD_TOKEN}}
```
stefanv, line 41: Add channel to this example, since default channel doesn't need this action.

README.md
```diff
@@ -18,11 +19,45 @@ jobs:
...
+```
+
+Which will remove all but the `n_latest_uploads` latest uploads from the channel. This is useful
+to avoid hosting outdated development versions, as well as to clean up space.
+
+The channel to remove wheels from is set to the ``scientific-python-nightly-wheels`` channel
+by default. If you are using this channel, please note that this repository will automatically
+clean up old artifacts for you.
```
stefanv, line 49:
```suggestion
Note that the ``scientific-python-nightly-wheels``, specifically, already removes
old artifacts daily. The `remove-wheels` action is therefore intended for use with other channels.
```

README.md
```diff
@@ -18,11 +19,45 @@ jobs:
...
+
+If you do not wish to have this automated cleanup, please open an issue on this repository
+to be added to the list of packages exempt from it. The current ones are named in 
+`packages-ignore-from-cleanup.txt`.
```
stefanv, line 53: I'd link to an explicit URL here, to make clear which repo we're talking about.

README.md
```diff
@@ -18,11 +19,45 @@ jobs:
...
+
+Please refer to the [artifact cleanup policy][] for more information.
+
+To remove wheels from a different channel, set the ``anaconda_nightly_upload_organization``
```
stefanv, line 57: Include this in the example instead.

README.md
```diff
@@ -45,7 +80,7 @@ then generate a token at `https://anaconda.org/<anaconda cloud user name>/settin
 with permissions to _Allow write access to the API site_ and _Allow uploads to Standard Python repositories_,
 and add the token as a secret to your GitHub repository.
 
-## Using a different channel
+## Using a different channel other than ``scientific-python-nightly-wheels``
```
stefanv, line 83:
```suggestion
## Using a channel other than ``scientific-python-nightly-wheels``
```

agriyakhetarpal: Oops, thanks!

remove-wheels/action.yml
```diff
@@ -0,0 +1,50 @@
...
+        locked: true
+        cache: true
+        cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+        # Avoid post cleanup errors if action run multiple times
+        post-cleanup: false
+
+    - name: Install dependencies
+      shell: bash
+      run: |
+        sudo apt-get update && sudo apt-get install -y curl jq
```
agriyakhetarpal, line 41: Is there a nicer way to add macOS support instead of having to do something like

```yaml
if: runner.os == 'macOS'
run: brew install curl jq # and so on
```

I don't know too much about how `pixi` works, yet – I probably should 😅

remove-wheels/action.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Remove old wheels
+      shell: bash
+      env:
+        INPUT_N_LATEST_UPLOADS: ${{ inputs.n_latest_uploads }}
+        INPUT_ANACONDA_USER: ${{ inputs.anaconda_user }}
+        INPUT_ANACONDA_TOKEN: ${{ inputs.anaconda_token }}
+      run: |
+        pixi run remove_old_wheels.sh
```
agriyakhetarpal, line 50:
```suggestion
      run: pixi run remove_old_wheels.sh
```

remove_wheels.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# get the anaconda token from the github secrets
+#
+# this is to prevent accidental removals
+echo "Getting anaconda token from github secrets..."
+
+ANACONDA_USER="${INPUT_ANACONDA_USER}"
+ANACONDA_TOKEN="${INPUT_ANACONDA_TOKEN}"
+N_LATEST_UPLOADS="${INPUT_N_LATEST_UPLOADS}"
+
```
agriyakhetarpal, line 20: I am not sure if the Anaconda token that is authorised to maintainers of, say, a package X is restricted to only remove wheels for said package X, or whether it can remove all packages in the index. This is because while users of SPNW won't be affected since we handle the deletions, users with their own organisation with wheels for packages X, Y, and Z being uploaded to it would either:

- run this action somewhere centrally to remove all (but the latest N) uploads of X, Y, and Z; or
- they would want to run the deletions from separate repositories for X, Y, and Z and want to delete just X, Y, and Z at a time on each package's own accord and deletion schedule.

Which situation would be more plausible? If it is the latter, we'd have to provide another input in the action for a comma-separated string of packages to delete uploads for (and maybe `*` for deleting all packages?). To replace the existence of `packages-ignore-from-cleanup.txt` (which would exist only in this repository – see below), it might also make sense to include a whitelist input, too. If it's the former, it gets easier to implement, but users won't have fine-grained control on automations for what packages are being deleted.

P.S. This is all valid only if I'm not missing something about how the index and its permissions are structured :) Happy to receive others' thoughts!

remove_wheels.sh
```diff
@@ -0,0 +1,109 @@
...
+# Query the package index for packages
+#
+# TODO: should be possible to alter this, since separating the workflow
+# into two steps, one for uploading and one for cleanup, should make it
+# possible for users to manually trigger the cleanup step before/after the
+# upload step has completed in their own repos instead of us having to do it.
+#
+# TODO: raises questions on how to moderate cleanups among multiple users
+# operating on the same channel, but that might be a different issue.
+curl https://raw.githubusercontent.com/scientific-python/upload-nightly-action/main/packages-ignore-from-cleanup.txt --output packages-ignore-from-cleanup.txt
```
agriyakhetarpal, line 48: Please see above for context and my question on this. :)

remove_wheels.sh
```diff
@@ -0,0 +1,109 @@
...
+if [ -s package-names.txt ]; then
+    threshold_date="$(date +%F -d '30 days ago')"
+
+    # Remember can't quote subshell as need to split on (space separated) token
+    for package_name in $(cat package-names.txt); do
+    # TODO: this outer loop can be removed when ready since there will be
+    # just one package to remove when the action is triggered manually from
+    # a user's (different) repo.
+
+        echo -e "\n# package: ${package_name}"
```
agriyakhetarpal, line 66: This TODO item would also be resolved if we have a way forward with the questions above, since, if multiple packages are removed at once, this loop can stay; if we were to restrict the action to removing just one wheel at a time (which I don't think we should), then we could remove this. I think an action of the form:

```yaml
- name: Remove old wheels
  uses: scientific-python/upload-nightly-action/remove-wheels@cantknowhashyet # 0.6.0
  with:
    n_latest_uploads: ${{ env.N_LATEST_UPLOADS }}
    anaconda_nightly_upload_organization: "your-organization"
    anaconda_nightly_token: ${{secrets.ANACONDA_TOKEN}}
    packages_to_remove: "mypackage1,mypackage2,mypackage3" # or just "*"
    # I could have suggested "all" here, but that breaks in the
    # case where "all" is also the name of a package that has
    # been uploaded (possible, albeit quite unlikely)
```

is better than specifying the step and authenticating multiple times.

agriyakhetarpal:
Thanks for the reviews so far! I can officially mark this ready for review now. I've added additional context about most of my questions in my review above.

tupui:
> I am not sure if the Anaconda token that is authorised to maintainers of, say, a package X is restricted to only remove wheels for said package X

There is a restriction. Users are only allowed to do actions on the packages they are given access to 😃 

There is just a handful of folks who have higher access right to administrate the Anaconda account. Here the token setup is very light but we use all the tools Anaconda gives us.

agriyakhetarpal:
Thanks for the clarification! Should we make a mention of this in the documentation?

tupui:
Feel free to make a suggestion 👍

remove-wheels/action.yml
agriyakhetarpal: I added `jq` to the dependencies in `pixi.toml` and removed this step, I guess that should work.

remove_wheels.sh
agriyakhetarpal: Based on @tupui's suggestion in https://github.com/scientific-python/upload-nightly-action/pull/95#issuecomment-2386447252, users will need to run the action multiple times to remove old uploads for multiple package versions. Hence, the second option of allowing per-package deletions is better, and an input for a list of packages to delete wheels for isn't required.

agriyakhetarpal: However, if someone has an appropriately scoped token for their organisation, they can very well remove multiple wheels from the index at a time, and some users might want to do that. So, it might be necessary to mention in the documentation how the tokens work and pre-emptively warn about possible deletions in an admonition.

remove_wheels.sh
agriyakhetarpal: I've put some thought into it, and this can be included as an input in the action. There would be two cases:

- An API token for package X that can delete older wheels for X
- An API token for the organisation that can perform index-wide deletions of older wheels for packages

While the former would be recommended to limit deletion scopes, when the latter is used by the action users (as it is being done here for this repo), a `whitelist_packages:` input can be added to include a comma-separated list of packages (`"openblas-libs"` in our case).

Another reasonable message could be to add a warning: `"Wheels for package X requested for deletion, but X is whitelisted. Please either remove it from the whitelist or try to delete a different package as needed."`.

remove_wheels.sh
tupui: To be clear, users have full control over their project. So they can add or remove any wheel they want on that project. People can just not add members to their project and they need to ask admins.

pixi.toml
```diff
@@ -11,3 +11,4 @@ version = "0.6.1"
 [dependencies]
 python = "3.12.*"
 anaconda-client = "==1.12.3"
+jq = "==1.7.1"
```
matthewfeickert, line 14: Please do not add additional dependencies to the default environment. This is the `pixi` manifest file for the upload action and additional dependencies should not be added to it.

I would recommend making a separate `pixi` project in a subdirectory.

Regardless, no dependency should be added without also updating a corresponding lock file.

README.md
```diff
@@ -1,7 +1,8 @@
 # Nightly upload
 
-This is a GitHub Action that uploads nightly builds to the [scientific-python nightly channel][],
-as recommended in [SPEC4 — Using and Creating Nightly Wheels][].
+This is a GitHub Action that uploads (and helps remove) nightly builds to
+the [scientific-python nightly channel][], as recommended in
+[SPEC4 — Using and Creating Nightly Wheels][].
```
matthewfeickert, line 5: These are two separate actions meant for very different audiences. Please do not combine them in the same areas of the README. They should be separated into different sections.

README.md
```diff
@@ -15,14 +16,46 @@ jobs:
       uses: scientific-python/upload-nightly-action@82396a2ed4269ba06c6b2988bb4fd568ef3c3d6b # 0.6.1
       with:
         artifacts_path: dist
-        anaconda_nightly_upload_token: ${{secrets.UPLOAD_TOKEN}}
+        anaconda_nightly_upload_token: ${{secrets.ANACONDA_TOKEN}}
```
matthewfeickert, line 19: Why is this getting changed?

matthewfeickert (changes requested):
I am in transit and so giving a quick but incomplete review. 

If the things I mention aren't clear please ask for clarification or more detail from me.

README.md
agriyakhetarpal: This is to suggest that since projects will use the same API key for their uploads and removals, having "UPLOAD" in the name (`UPLOAD_TOKEN`) wouldn't make sense for removals. Is there a better way to convey this information?

agriyakhetarpal: Oh, your comment above of separating the actions into different sections should fix this, let me revert this change.

pixi.toml
agriyakhetarpal: Thanks for this – I added a separate manifest at `remove-wheels/pixi.toml` like you requested in 4a3497d3b81ce7eee22d4b632fd3d6471377b9de and added a lockfile for it in 71ede798fffb686daf5bb5d49096af579859ac06.

matthewfeickert: > I'm assuming that the lockfile is cross-platform

@agriyakhetarpal Yes, that is what the 

```toml
platforms = ["linux-64", "osx-64", "osx-arm64"]
```

that you added does for you. :+1: 


The `pixi` docs are very useful and approachable: https://pixi.sh/ So I'd recommend a speed read via searching it for key words.

agriyakhetarpal: Awesome! Yes, I was skimming the docs in the meantime – they're neat :)

README.md
```diff
@@ -18,11 +18,43 @@ jobs:
...
 ```
 
-Note that we recommend pinning the action against a specific SHA
+> [!IMPORTANT]
+> Note that we recommend pinning the action against a specific SHA
 (rather than a tag), to guard against the unlikely event of upstream
 being compromised.
 
-## Updating the action
+# Removing old nightly builds
```
stefanv, line 26:
```suggestion
## Removing old nightly builds
```

stefanv: Presuming `#` is kept fot title?

README.md
```diff
@@ -18,11 +18,43 @@ jobs:
...
+
+```yml
+jobs:
+  steps:
+    ...
+    - name: Remove old wheels
+      uses: scientific-python/upload-nightly-action/remove-wheels@cantknowhashyet # 0.6.0
+      with:
+        n_latest_uploads: ${{ env.N_LATEST_UPLOADS }}
+        anaconda_nightly_upload_organization: "your-organization"
```
stefanv, line 40: This is called organization, not channel; are these interchangeable terms? If so, use channel; if not, explain the difference.

README.md
```diff
@@ -18,11 +18,43 @@ jobs:
...
+        anaconda_nightly_token: ${{secrets.ANACONDA_TOKEN}}
+```
+
+Which will remove all but the `n_latest_uploads` latest uploads from the channel. This is useful
+to avoid hosting outdated development versions, as well as to clean up space.
+
+Note that the ``scientific-python-nightly-wheels`` channel, specifically, already removes
```
stefanv, line 47: Should we move this up, so the person doesn't read all of this, make a change to their CI, and only then read the part where it's unnecessary?

remove_wheels.sh
stefanv: `packages: ["a", "b", "c"]` hopefully!

stefanv (approved):
Looks good to me overall; I made two more suggestions for clarifying the documentation.

README.md
agriyakhetarpal: Oh, I was keeping this because this repository would support two major actions. I'll keep just one title for the README.

remove_wheels.sh
agriyakhetarpal: Yes, a list of strings sounds elegant! :)

README.md
agriyakhetarpal: Addressed in 556e9a6bca0f2d1ff406804ed59829400c6a9fbe.

README.md
agriyakhetarpal: Addressed in e403c81ce80957498a90530b41b9d19ac6ab16f2.

agriyakhetarpal: Yes, that's reasonable – addressed in e403c81ce80957498a90530b41b9d19ac6ab16f2.

README.md
agriyakhetarpal: I'm not quite sure, since they look like they have been used interchangeably. For example, the README outside of this PR:

https://github.com/scientific-python/upload-nightly-action/blob/920fb594ddba437d939052cabd2a59ceb67df265/README.md?plain=1#L48-L67

mentions how one may upload to a different channel, but it uses "organization" as an input.

agriyakhetarpal:
CI is going to fail here because I don't have an `ANACONDA_TOKEN` set up in my fork's secrets. I could set an Anaconda organisation up (instructions would be welcome) for testing (or look into failures in a new PR, though I don't recommend that).

Or, the best way I can think of is that I could try this with a project in a PR downstream with the action set to my fork – since I help co-maintain PyWavelets's nightly wheels as I mentioned earlier up somewhere :) I would be unavailable next week, please let me know if this is reasonable and I'll try it out as early as the week after.

Cadair:
I would be good if this could get merged, it would mean we could replace https://github.com/OpenAstronomy/publish-wheels-anaconda with this action in a few places.

agriyakhetarpal:
I'll work towards resolving the conflicts while we are all here in person. However, it might take me some time to remember what I was doing here and go through the past discussions :) The last thing I remember working on was doing a smoke test with an Anaconda index of my own, but I don't think I got around to it on my own.

agriyakhetarpal:
@Cadair and I have modified the action from this repository's workflows to run in a way that deletes packages one at a time (through a dynamically created matrix based on the available wheels in the specified index), rather than deleting multiple packages at once.

bsipocz:
> is still the plan/goal of this PR I think that should be avoided. It is a good idea and fine to give projects the ability to clean up their own wheels easily, but that ability should not be removed from the org by fully delegating all abilities.

Fully agree, we should still be able to curate the packages, too, I mean the ones we host (though I believe Cadair's use case is to use all these machinery to upload to elsewhere  -- and their the curation would fall on us)

.github/workflows/remove-wheels.yml
```diff
@@ -20,15 +20,70 @@ env:
...
+        run: env
+
+      - name: Show CLI API info
+        run: |
+          anaconda show --help
+          echo ""
+          anaconda remove --help
+
+
+      - name: Generate list of package names
```
stefanv, line 36: Is this package list generation still necessary?

agriyakhetarpal: `get-pkgs` generates the input matrix, so yes

agriyakhetarpal:
(CI fails because this PR is from my fork)

bsipocz:
> (CI fails because this PR is from my fork)

Could you rebase to resolve the conflicts?

agriyakhetarpal:
Thanks! Done. I also renamed `n_latest_uploads` to `n_latest_uploads_to_keep` as that sounds clearer to me.

agriyakhetarpal:
Turns out `jq` isn't on Windows yet (https://github.com/conda-forge/jq-feedstock/issues/3, https://github.com/conda-forge/jq-feedstock/pull/24), which is surprising – this kind of explains why you were wanting me to rewrite the wheel removal script in Python instead, @Cadair 😅

.github/workflows/ci.yml
```diff
@@ -96,13 +96,27 @@ jobs:
...
+      - name: Remove test package uploads via the action
+        uses: ./remove-wheels
+        with:
+          n_latest_uploads_to_keep: 0
+          anaconda_nightly_upload_organization: scientific-python-nightly-wheels
+          anaconda_nightly_token: ${{ secrets.ANACONDA_TOKEN }}
+          package_name: test-package
+
+      # The action removes per version, so delete the (now empty) package
+      # registration as well
```
agriyakhetarpal, line 112: I added this in the last commit in 7f6204c48c29b064872139bfa78aec504d457514 as a means to test the action on PRs (otherwise it runs only on a daily schedule and on manual triggers). Please let me know in case you think this is not ideal – I'll revert. Thanks!

stefanv (approved):
This looks like a great improvement. Thanks @agriyakhetarpal!

Issue #150, opened 2025-12-10
Support uploading windows and linux aarch64 wheels


mroeschke:
Over in pandas we're trying to use this action to upload to `https://anaconda.org/channels/scientific-python-nightly-wheels/packages/pandas/overview` and the following platforms are failing (example failing job: https://github.com/pandas-dev/pandas/actions/runs/20086600506)

* win-64
* win-arm64
* manylinux-aarch64
* musllinux-aarch64

I think these are largely failing because these platforms are not specified in the `platforms` in the pixi.toml https://github.com/scientific-python/upload-nightly-action/blob/64b7bfd3a1e0e823923208401112dd82dd8ceef5/pixi.toml#L6

matthewfeickert:
You're correct, though the assumption is that most people are collating builds and uploading them in a single job (e.g. [Matplotlib's workflow](https://github.com/matplotlib/matplotlib/blob/fdf8995b539e8dc93a427e0eef26068b7a2b42af/.github/workflows/nightlies.yml)).

Your PR is probably reasonable to support people however they would like, but is there a reason you're uploading individually, instead of uploading only on verified success of all builds?

matthewfeickert:
> ... and the following platforms are failing (example failing job: https://github.com/pandas-dev/pandas/actions/runs/20086600506)
> ...
> * win-arm64


Given the lack of `win-arm64` supported feedstocks on conda-forge (c.f. https://github.com/scientific-python/upload-nightly-action/pull/151#pullrequestreview-3566015421) you will need to do some consolidation of your upload process to get the `win-arm64` wheels uploaded.

PR #151, opened 2025-12-10
ENH: Add aarch64 Linux and x86 Windows support

mroeschke:
closes https://github.com/scientific-python/upload-nightly-action/issues/150

I also tried adding `win-arm64` but the `pixi lock` fails with

```bash
$ pixi lock
Error:   × failed to solve requirements of environment 'default' for platform 'win-arm64'
  ├─▶   × failed to solve the environment
  │   
  ╰─▶ Cannot solve the request because of: No candidates were found for python 3.13.*.
```

I tried unpinning python to `3.*`, and it still cannot solve, so this may be a pixi or window-arm ecosystem support issue

**Squash and merge commit message:**

```
* Add linux-aarch64 and win-64 platforms.
* Relock lock file.
```

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -9,101 +9,101 @@ environments:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
-      - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-cli-base-0.7.0-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-cli-base-0.5.2-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-client-1.13.0-pyh29332c3_1.conda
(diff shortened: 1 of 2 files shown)


matthewfeickert (approved):
I think that this addition is fine. We can do some additional small updates and then make a release so that if other people want to get this while staying pinned to a release version they can.

> I also tried adding win-arm64 but the pixi lock fails with
> ...
> I tried unpinning python to 3.*, and it still cannot solve, so this may be a pixi or window-arm ecosystem support issue

It is neither really, but kinda the Windows one. It is just that `win-arm64` is relatively new on conda-forge AFAIK (though I can't find a conda-forge blog post on it, so I'm not sure how long there has been preliminary support). You might know more about the history here than me, but in 2025 there are still very few conda-forge feedstocks with `win-arm64` support, and [`anaconda-client`](https://github.com/conda-forge/anaconda-client-feedstock/commit/89047a6c5a7983bced04500d30012a8e221da2fa) is not one of them. So until there are `win-arm64` builds for [all of its dependencies](https://github.com/conda-forge/anaconda-client-feedstock/blob/89047a6c5a7983bced04500d30012a8e221da2fa/recipe/recipe.yaml#L25-L52) we can't do anything about that.

**edit:** Oh, I guess [Python itself also does not have `win-arm64`](https://github.com/conda-forge/python-feedstock/tree/c90e3cfe0a44db1b0741359d68c330cd7d861b16) support yet on conda-forge.

matthewfeickert:
:wave: Ping @scientific-python/nightly-wheels-developers for another review on this PR beyond me.

bsipocz (approved):

PR #162, opened 2026-04-20
ENH: Add month cooldown window for new package uploads

matthewfeickert:
* To weakly guard against security issues that are introduced in new uploads to conda-forge, add a 30 day exclusion window for Pixi's resolver.
   - [SP Discord thread that got me thinking about this](https://discord.com/channels/786703927705862175/1257384941990318121/1493960401992548453)
* This feature was introduced in Pixi v0.67.0, so require at least that for use.
* Update anaconda-client to v1.14.1.

This PR is coming from the main project and **not** my fork so that the CI actually runs, which is important on `anaconda-client` updates.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -9,10 +9,13 @@ environments:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-auth-0.12.3-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-cli-base-0.7.0-pyhd8ed1ab_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-client-1.13.1-pyhcf101f3_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anaconda-client-1.14.1-pyhcf101f3_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/annotated-types-0.7.0-pyhd8ed1ab_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/attrs-25.4.0-pyhcf101f3_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/backports-1.0-pyhd8ed1ab_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/backports.tarfile-1.2.0-pyhcf101f3_2.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/backports.zstd-1.2.0-py314h680f03e_0.conda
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@scientific-python/nightly-wheels-developers this is ready for review.

hugovk:
> * To weakly guard against security issues that are introduced in new uploads to conda-forge, add a 30 day exclusion window for Pixi's resolver.

On the contrary, this can be very effective.

For example, of the 10 attacks in the chart at https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns, seven were taken down in under 24 hours. The others were 3 and 10 days, and 5 weeks.

lucascolley (approved):
thanks Matthew!

tupui (approved):

matthewfeickert:
Thanks for the fast review, all!

> On the contrary, this can be very effective.

I agree @hugovk. My thoughts are more along this section from the blog

> Cooldowns are, obviously, **not a panacea**: some attackers _will_ evade detection, and delaying the inclusion of potentially malicious dependencies by a week (or two) does not fundamentally alter the fact that supply chain security is a _social trust_ problem, not a purely technical one. Still, an 80-90% reduction in exposure through a technique that is free and easy seems hard to beat.

My main concern and motivating reason for "weakly" is that we're relying on both detection to have happened and for this to be reported and properly dealt with by conda-forge. I have no doubt that `conda-forge/core` takes security seriously ([they obviously do](https://conda-forge.org/blog/2025/04/02/security-incident-with-package-uploads/)), but they do emphasize 

> conda-forge packages are built by strangers on the internet (our wonderful feedstock maintainers!) and are not suitable for use cases that require secure software provenance.

and we currently have no way of performing automated security audits of software environments if conda packages are installed (c.f. [#general > ✔ Recommendations for security audits on conda packages? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.E2.9C.94.20Recommendations.20for.20security.20audits.20on.20conda.20packages.3F/near/585660514)). I think that until we can overcome this last issue that we can't really say more than "weakly", even if we're doing our best.

henryiii:
Why a month? Generally 3 or 7 days is what I've seen projects adopt. At a month, that's quite a bit of time to also be isolated from security updates.

hugovk:
We use 14 days for CPython.

How and when do you update pixi.lock?

I know Renovate supports pixi.lock, does Dependabot? If you use those, they ignore the schedule and cooldown for security updates.

matthewfeickert:
> Why a month?

@henryiii Not all conda-forge feedstocks are as responsive as their upstream PyPI or source components. A month gives sufficient time for things to work their way through the system. I'm not against changing this though.

> How and when do you update pixi.lock?

@hugovk We've been [manually updating it as needed](https://github.com/scientific-python/upload-nightly-action/commits/main/pixi.lock), generally motivated by updates to https://github.com/anaconda/anaconda-client

https://github.com/scientific-python/upload-nightly-action/blob/7c719ad65b60aa6c99622ddf4135b487d87767ed/pixi.toml#L15

which are pretty infrequent.

We recommend that users are locked to a particular commit hash instead of targeting the workflow repository `HEAD`.

https://github.com/scientific-python/upload-nightly-action/blob/7c719ad65b60aa6c99622ddf4135b487d87767ed/README.md#L10-L23

Issue #171, opened 2026-07-22
Re-opening the discussion about retention policy


betatim:
(This is related to https://github.com/scientific-python/upload-nightly-action/issues/119 but that issue is closed and quite old so I thought I'd make a new one)

[scikit-learn recently ran into the problem](https://github.com/scikit-learn/scikit-learn/pull/34528) that we didn't upload a nightly wheel for >30 days. As a result there were no wheels in the nightly channel which meant the testing with nightly wheels for a different package broke.

I thought/understood the [policy in the README](https://github.com/scientific-python/upload-nightly-action/tree/main#artifact-cleanup-policy-at-the-scientific-python-nightly-wheels-channel) as "we will keep the last five wheels no matter what, everything else >30 days gets deleted".

While I agree that a wheel >30days old is no longer a "nightly" really, it also seems annoying to break the CI of a package that uses the wheels for no other reason than that we didn't manage to fix our nightly wheel producer job in a timely fashion. The problem in this case was not not noticing, just not being fast enough.

I'd propose we keep the last five wheels no matter what.

Do we want to re-open the discussion?

Carreau:
if prefer a signal that something is broken (no more wheels) than silently not testing nightly.

To disagree with myself (because of course) it is hard to know which project are supposed be available once the last wheel is removed. 

Maybe we can strike a mid-ground; if 30 days is reached; open an issue and give a grace period of 30days after which the issue is closed and last wheel deleted ?

bsipocz:
I like the middle ground suggestion better as I vaguely recall a situation in the past with terribly outdated nightlies (that were older than the latest release), something us downsteam users didn't notice and just assumed we had a passing dev testing build. And not having a strictly time-based cleanup mechanism, in theory, we could end up in the same situation here.

bsipocz:
OTOH, I think there is value in breaking downstream in the situation above, as they are indeed not testing the latest dev version. It's not their fault, but still, they don't have a passing build and should expect possible problems when a new nightly is out -- a heads up is thus great.

The situation in  #119 is different in the sense that in smaller packages there may be legit 30day gaps without commits, so the old nightly could be still the most up-to-date.

matthewfeickert:
I haven't fully caught up on the conversation (sorry), but just jumping in to ask what's the problem with having projects just upload on a regular cadence even if they don't have new changes? Does

[...]

PR #173, opened 2026-07-27
Add details about the consequences of not regularly uploading wheels

betatim:
This is a follow up to https://github.com/scientific-python/upload-nightly-action/issues/171 that adds some details to the README to help low traffic projects. It highlights the problem that the nightly wheels will get cleaned up if you don't produce a wheel regularly and that this is something to consider for low traffic projects.

I also wrote that "we are not aware of a way to prevent the cron job expiry", mostly because I was not aware of a solution. There is `.github/workflows/keep-alive.yml` which I think is a solution to this problem. Should we advertise that as a potential solution?

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,16 @@ Any versions beyond these are automatically removed as part of a daily cron job 
 Projects may have reasons to request to be added to the list exempt from this automated cleanup, however
 in that case the responsibility of cleaning-up old, unused versions fall back on the individual project.
 
+If you are a low traffic project this policy might catch you out, as the last wheel will be removed from
+this channel 30 days after it was uploaded. A low activity project might not receive updates often enough
+to produce a new wheel every 30 days. We recommend that you upload a nightly wheel on a regular cadence
+even if there are no changes.
+
+In addition cron jobs in GitHub repositories will be disabled after a certain amount of inactivity. We
+are not aware of a solution other than regular commit activity to prevent the deactivation of scheduled
+cron jobs.
+
+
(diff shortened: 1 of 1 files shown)


Carreau (approved):

README.md
```diff
@@ -77,6 +77,16 @@ Any versions beyond these are automatically removed as part of a daily cron job
...
 in that case the responsibility of cleaning-up old, unused versions fall back on the individual project.
 
+If you are a low traffic project this policy might catch you out, as the last wheel will be removed from
+this channel 30 days after it was uploaded. A low activity project might not receive updates often enough
+to produce a new wheel every 30 days. We recommend that you upload a nightly wheel on a regular cadence
+even if there are no changes.
+
+In addition cron jobs in GitHub repositories will be disabled after a certain amount of inactivity. We
+are not aware of a solution other than regular commit activity to prevent the deactivation of scheduled
+cron jobs.
```
matthewfeickert: To avoid churn on the diffs of the text can you break at the end of each sentence, like the rest of the README does?

matthewfeickert (commented):
Content looks good. Small suggestion on text maintenance.

README.md
betatim: TIL!

Done

matthewfeickert (approved):
LGTM

I'm going to leave this open for one more day so that @bsipocz can leave any comments that she might have, but anyone should feel free to merge this at this point IMO.

bsipocz (approved):
Looks good, thank you!

PR #175, opened 2026-08-13
Add issue-opener when wheels are almost dead

### LINKED ISSUE #171: Re-opening the discussion about retention policy

larsoner:
Drafted by Claude Opus 5, reviewed by me. Retention policy unchanged. But now with this action run in this repo on a daily schedule (1h before wheel removal):

1. Once a wheel is 15 days old, open an issue in the repo responsible for the wheel (pulled from PyPI info)
2. At 25 days, comment on the issue.
3. Once a new wheel is uploaded, close the issue if one is open.
4. Open an issue here if for some reason we can't on the source repo.
5. Open an issue here if the action itself fails (to make sure we don't silently fail!)
6. Issues will be opened by https://github.com/scientific-python-bot , a bot user I just created with the scientific-python logo (so it'll be posting as us / our org), and added a README for so people can understand what that user is
7. If a maintainer closes an issue instead of us, stay quiet for 60 days.

Eventually if for some reason this becomes annoying, we can add an opt-out list, different reporting stages, etc. But I think this is a good start. Comes with some minimal tests. Ended up being a bit longer than I expected, but it covers a bit more ground/cases than I realized it should originally.

Closes #171

Once this is merged I'll `workflow_dispatch` once in dry-run mode, then once that's good once again in full mode. Both of these should say everything worked okay and that it's acting as the `scientific-python-bot`. And I expect issues for `sunpy` (29 days, hours from losing its wheels), `pyproj` (19), and `scikit-image` (18).


larsoner:
The [CI / test upload via action (pull_request)](https://github.com/scientific-python/upload-nightly-action/actions/runs/31721324170/job/94518743961?pr=175) is red on every PR, which is not good. I'll push a little commit for that, too.

larsoner:
... and since this was largely agent-authored, I'll push a commit to seed an AGENTS.md (with a CLAUDE.md pointing to it) as well so future agent-based PRs are easier. Could be a separate PR (or omitted entirely!), but seems reasonable to add it here.

tests/test_check_stale_wheels.py.lock
bsipocz, line 1: why do we need these lock files?

larsoner: To fix the dep pinning. But I think it'll be better to use a `requirements.txt` and let dependabot update it... I'll push that instead.

larsoner:
@bsipocz good to go now? I think this is the sort of thing it might be easiest to try and then refine iteratively if it doesn't work

larsoner:
Not sure why I was assigned @matthewfeickert ? This is my PR and I've addressed review comments... 🤔

matthewfeickert:
> This is my PR

This is how PR assignment works. https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/assigning-issues-and-pull-requests-to-other-github-users

I'm not sure what is unclear about this. Do you use a different standard in your projects?

larsoner:
Yeah I don't remember ever being assigned a PR that I've authored (so if it has happened, I think it's rare), just issues.

So I'm not quite sure what it signifies... not sure what I (or others) are supposed to take away, understand, do, etc. from me being assigned to my PR. Does it just mean it's mine to work on and edit and others shouldn't or something? That's at least what I've understood issue assignment means. But I figured for a PR, that's typically the default state: leave it for the PR author to work on (other than maybe minor maintainer tweaks before merge), hence my confusion...

larsoner:
Okay for you @bsipocz ? Would be good to start trying this out before another wheel bites the dust!

matthewfeickert:
Maybe this is more common in larger orgs? To me this is quite normal; every PR gets assigned for clarity.

> not sure what I (or others) are supposed to take away, understand, do, etc. from me being assigned to my PR.

You can group by assignment. As I linked above, assignment makes it clear who is the responsible party/parties for implementation, which is why you can have multiple people assigned to a PR.

> But I figured for a PR, that's typically the default state: leave it for the PR author to work on (other than maybe minor maintainer tweaks before merge), hence my confusion...

Indeed, the default assignment is to the PR creator, but if responsibility shifts it is useful to formalize this. As I said, every Issue and every PR gets assigned (ideally) so that there's any hope at being able to track states by team members.

---

> Would be good to start trying this out before another wheel bites the dust!

@larsoner I'm currently about to go into meetings but I have some revision suggestions that I'll leave in the next few hours.

.github/workflows/report-failure.yml
```diff
@@ -0,0 +1,42 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TITLE: ${{ inputs.title }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          body="\`${GITHUB_WORKFLOW}\` failed, most recently in [this run](${RUN_URL}).
+
+          Until it is fixed, whatever that workflow keeps an eye on is unattended."
+
+          number=$(gh issue list --repo "${GITHUB_REPOSITORY}" --state open --json number,title \
```
matthewfeickert: The default limit is 30

```console
$ gh issue list --help

...

FLAGS
...
  -L, --limit int          Maximum number of issues to fetch (default 30)
```

which could realistically be below the open issue count on some repos. Bump this to something larger like 200 or more? (`gh` paginates at 100 per page, so a few API calls isn't that bad)

```suggestion
          number=$(gh issue list --repo "${GITHUB_REPOSITORY}" --state open --limit 200 \
              --json number,title \
```

matthewfeickert (changes requested):
RE: https://github.com/scientific-python/upload-nightly-action/pull/175#issuecomment-5343141830, this is one of a few. I'll leave more in a few hours.

.github/workflows/report-failure.yml
larsoner: Good call, fixed -- but by using `--author`. FWIW this part of the action is for opening issues on this repo when its own actions fail for some reason (i.e., not for opening issues on repos that upload wheels -- that's a separate step done elsewhere). We currently have 20 issues open total, and none from the bot, so with `--author` the default limit of 30 should be fine for finding non-stale issues opened by the bot I think.

The issue count for other repos like sklearn etc. is handled down in `tools/check_stale_wheels.py`, and it uses `login=...` which limits it to issues opened by the scientific-python bot, so I think we're hopefully good there.

bsipocz:
> Okay for you @bsipocz ? Would be good to start trying this out before another wheel bites the dust!

Yeap, looking at it now.

As for PR assignments, I also feel that's an overkill and we don't do that in community OSS. We don't/should not really do review assignments either besides pings -- it's different when someone self-assigns, but as none of us are each others manager we shouldn't really assign anything, unless that person's review/sign-off is an actual blocker. IMO.

.github/workflows/report-failure.yml
bsipocz, line 1: I think this will be a nice short action to reuse elsewhere, too. So ultimately, I think it would make sense to actually document/mention all of the actions, suitable for direct downstream usage, in the README.

.github/workflows/stale-wheels.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9  # v9.0.0
+        with:
+          python-version: '3.14'
+
+      - name: Check for stale wheels
+        env:
+          ISSUE_OPENER_TOKEN: ${{ secrets.ISSUE_OPENER_TOKEN }}
```
bsipocz, line 37: Do we have this token yet?

.github/dependabot.yml
```diff
@@ -11,3 +11,15 @@ updates:
           - "*"
     labels:
       - "github-actions"
```
bsipocz, line 13: Let's add this for consistency.
```suggestion
      - "github-actions"
      - "chore"
      
```

README.md
```diff
@@ -86,6 +86,18 @@ In addition cron jobs in GitHub repositories will be disabled after a certain am
 We are not aware of a solution other than regular commit activity to prevent the deactivation of
 scheduled cron jobs.
 
+### Stale wheel reminders
```
bsipocz: Maybe this should be the same heading level as the rest? Not a strong prefernce

```suggestion
## Stale wheel reminders
```

AGENTS.md
bsipocz, line 1: I haven't reviewed or even read this file.

tools/check_stale_wheels.py
```diff
@@ -0,0 +1,465 @@
+#!/usr/bin/env python
+"""Report projects whose nightly wheels are about to be removed from the channel.
+
+Wheels older than the retention period are deleted by ``remove-wheels.yml``, which
+can leave a project with no nightly wheels at all and break downstream CI. This
+script runs on a schedule from this repository, finds packages whose most recent
+upload is older than the warning threshold, and opens an issue on the project's
+own tracker so the maintainers hear about it before the wheels disappear.
+
+Requires ``ISSUE_OPENER_TOKEN`` (a classic token with the ``public_repo`` scope,
```
bsipocz, line 10: I think this docs should be exposed in the README.

tools/check_stale_wheels.py
```diff
@@ -0,0 +1,465 @@
...
+CHANNEL_URL = f"https://anaconda.org/{ANACONDA_USER}"
+ACTION_URL = "https://github.com/scientific-python/upload-nightly-action"
+POLICY_URL = f"{ACTION_URL}#artifact-cleanup-policy-at-the-scientific-python-nightly-wheels-channel"
+
+# Hidden markers let us find our own issues and comments again without relying on
+# the (fuzzy, eventually consistent) search API.
+MARKER = "<!-- scientific-python-nightly-wheels: stale-wheels -->"
+FINAL_WARNING_MARKER = "<!-- scientific-python-nightly-wheels: stale-wheels-final -->"
+REPORT_MARKER = "<!-- scientific-python-nightly-wheels: stale-wheels-report -->"
+
```
bsipocz, line 43: nor a merge blocker, but I wonder if things fail loudly if any of these markers or URLs change/stop working?

README.md
```diff
@@ -86,6 +86,18 @@ In addition cron jobs in GitHub repositories will be disabled after a certain am
...
+
+So that the removal above does not come as a surprise, a daily job in this repository checks how
+long ago each package was last uploaded. If it has been more than **15 days**, the job opens an
+issue on that project's own issue tracker naming the date its wheels will be removed, comments once
+more at 25 days, and closes the issue automatically once a new wheel is uploaded. Issues are opened
```
bsipocz: consistency, though I better like how we do it in the previous section.

```suggestion
more at **25 days**, and closes the issue automatically once a new wheel is uploaded. Issues are opened
```

tools/check_stale_wheels.py
```diff
@@ -0,0 +1,465 @@
...
+        if not os.environ.get(name):
+            parser.error(f"{name} is not set")
+
+    GH = Github(auth=Auth.Token(os.environ["ISSUE_OPENER_TOKEN"]))
+    # An unusable token (usually expired) fails the run, which the workflow turns
+    # into an issue; report_errors only covers projects we could not reach.
+    LOGIN = GH.get_user().login
+    print(f"Acting as {LOGIN}")
+
+    names = list_packages(read_ignored(IGNORE_FILE))
```
bsipocz: I'm confused, shouldn't this be all but the ignored packages, so shouldn't this be the complimenter?

tools/check_stale_wheels.py
```diff
@@ -0,0 +1,465 @@
...
+and downstream projects that test against nightlies will fail to install them.
+
+Some things worth checking:
+
+- Is the workflow that builds and uploads the nightly wheels failing?
+- Was its scheduled run \
+[disabled by GitHub](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-a-workflow) \
+after a period of repository inactivity?
+- Has there simply been nothing to build? Uploading on a fixed cadence even when nothing has \
+changed keeps the channel populated for downstream users.
```
bsipocz:
```suggestion
- Has there simply been nothing to build? Consider uploading on a fixed cadence even when nothing has \
changed to keep the channel populated for downstream users.
```

bsipocz (changes requested):
OK, so I have a bunch of nitpicky wordsmithing and documentation related comments. None of them are blockers.

However, please address my question about the packages included in the loop. IMO, that's the only blocker, but it maybe just a comment that resolves the confusion, or maybe a fix if I read things correctly.

.github/workflows/stale-wheels.yml
larsoner: Yep I added it!

larsoner:
Okay, taken care of @bsipocz! Took/implemented the suggestions.

On failing loudly, the idea is that anything that stops us reaching a project's tracker (no usable URL, issues disabled, bad token, etc.) gets collected into a single issue in this repo that's rewritten each run, and a failure of the workflow itself opens one here too, so it should always signal needing manual intervention. The hidden markers are the one thing that can't self-report: if we renamed one, the issues already carrying it would be orphaned and we'd open new ones alongside them, so there's a comment in the script saying not to do so. A maintainer stripping it out of an issue body just gets them a fresh issue, which seems fine.

bsipocz (approved):
Thanks, it's looking good now and I agree that we can iterate on it in case we need adjustments and follow-ups.

larsoner:
It would be great to merge this in the next day if possible. A few wheels are near expiry. (Can open issues manually of course, but they would be an excellent test of the mechanics and a good opportunity for ironing out any residual bugs!)

bsipocz:
OK, given Matthew's thumbs up on your response to his blocking review, I would go ahead and consider it addressed and merge this without waiting for another round from him.

larsoner:
Okay I'll `workflow_dispatch` and :crossed_fingers: :sweat_smile:

larsoner:
Dry run [looking good](https://github.com/scientific-python/upload-nightly-action/actions/runs/32417411097):

image

So I ran it in non-dry-run mode and it [opened the three issues](https://github.com/scientific-python/upload-nightly-action/actions/runs/32418578541) :rocket:

bsipocz:
I really love how good the report summary looks like!

matthewfeickert:
Sorry for getting fully taken away by some work critical things, but thanks a whole bunch for this work @larsoner! I agree with @bsipocz that this is super slick to have a nice table view like https://github.com/scientific-python/upload-nightly-action/actions/runs/33702171407. :)

I guess future requests would be to have this get published as a status page website so that we have a single URL to go and look at how things stand.

bsipocz:
> I guess future requests would be to have this get published as a status page website so that we have a single URL to go and look at how things stand.

I suppose we could just push it to this repo's github pages?

